PR #1, opened 2022-05-16
Initialization of project

marcin-jozefowicz:


README.md
```diff
@@ -0,0 +1,21 @@
+## Quick start
+
+### Backend: PostgreSQL & API
+
+You can spin up DB one easily using docker:
+
+```sh
+# use "adopttapir" as a password
+docker run --name bootzooka-postgres -p 5432:5432 -e POSTGRES_PASSWORD=adopttapir -e POSTGRES_DB=adopttapir -d postgres
```
adamw, line 9: do we need a database? :)

backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterApi.scala
```diff
@@ -0,0 +1,61 @@
...
+import sttp.model.StatusCode
+
+class StarterApi(http: Http, starterService: StarterService, xa: Transactor[IO]) {
+  import http._
+
+  private val starterPath = "starter.zip"
+
+  private val starterEndpoint = baseEndpoint.get
+    .in(starterPath)
+    .in(query[String](tapirVersion))
```
adamw, line 19: maybe modelling this as a json body would give us more flexibility? if we introduce more parameters at some point ...

adamw (approved):

README.md
marcin-jozefowicz: Indeed we don't need it for now.

backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterApi.scala
marcin-jozefowicz: Sure good idea!

PR #16, opened 2022-06-01
Support for optional documentation endpoints.

### LINKED ISSUE #5: Add support for documentation endpoints
Introduction in #4 

Client can choose if he want to have documentation endpoints. (Yes/No) 
 - Add support for documentation endpoints.

marcin-jozefowicz:
closes #5

### DIFF
--- a/backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterModel.scala
+++ b/backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterModel.scala
@@ -6,7 +6,7 @@ case class StarterDetails(
     serverEffect: ServerEffect,
     serverImplementation: ServerImplementation,
     tapirVersion: String,
-    documentationAdded: Boolean
+    addDocumentation: Boolean
 )
 
 object StarterDetails {
--- a/backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterModel.scala
+++ b/backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterModel.scala
@@ -5,7 +5,8 @@ case class StarterDetails(
     groupId: String,
     serverEffect: ServerEffect,
     serverImplementation: ServerImplementation,
-    tapirVersion: String
+    tapirVersion: String,
+    documentationAdded: Boolean
 )
 
 object StarterDetails {
--- a/backend/src/main/scala/com/softwaremill/adopttapir/template/scala/APIDefinitionsView.scala
+++ b/backend/src/main/scala/com/softwaremill/adopttapir/template/scala/APIDefinitionsView.scala
@@ -10,6 +10,13 @@ object APIDefinitionsView {
     case ServerEffect.ZIOEffect    => HelloServerEndpoint.zio
   }
 
+  def getDocEndpoints(starterDetails: StarterDetails): PlainLogicWithImports = {
+    if (starterDetails.documentationAdded) {
+      DocumentationEndpoint.prepareDocEndpoints(starterDetails.projectName, starterDetails.serverEffect)
+    } else
+      PlainLogicWithImports.empty
+  }
+
   private object HelloServerEndpoint {
 
     val future: PlainLogicWithImports = PlainLogicWithImports(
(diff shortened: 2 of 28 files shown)


backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterModel.scala
```diff
@@ -5,7 +5,8 @@ case class StarterDetails(
     groupId: String,
     serverEffect: ServerEffect,
     serverImplementation: ServerImplementation,
-    tapirVersion: String
+    tapirVersion: String,
+    documentationAdded: Boolean
```
adamw, line 9: [minor, but naming] -> `addDocumentation`? it's not yet added ;)

backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITSpec.scala
```diff
@@ -119,7 +123,138 @@ class StarterServiceITSpec extends BaseTest {
...
+      .unsafeRunSync()
+  }
+
+  it should "return zip file containing working sbt folder with Future Akka implementation with doc endpoint" in {
+    val service = createStarterService
+
+    val starterDetails =
+      StarterDetails("projectName", "com.softwaremill", FutureEffect, Akka, defaultTapirVersion, documentationAdded = true)
+
+    IO.blocking(BFile.newTemporaryDirectory("sbtTesting"))
```
adamw, line 151: [minor] can be probably extracted to a helper method :)

adamw: if the tests are the same but vary in parameters then we should extrat the common parts, much easier to read

backend/src/main/twirl/MainFutureAkka.scala.txt
```diff
@@ -13,7 +16,7 @@ object Main {
   def main(args: Array[String]): Unit = {
     implicit val actorSystem: ActorSystem = ActorSystem()
 
-    val route = AkkaHttpServerInterpreter().toRoute(List(ApiDefinitions.helloServerEndpoint))
+    val route = AkkaHttpServerInterpreter().toRoute(ApiDefinitions.helloServerEndpoint@if(documentationAdded) { :: ApiDefinitions.docEndpoints})
```
adamw, line 19: Maybe it would be cleaner to rename `ApiDefinitions` -> `Endpoints`, and there have:
```
object Endpoints {
  val hello = ...
  val docs = ...
  val all = List(hello) ++ docs
}
```

and then reference `Endpoints.all` here ?

backend/src/main/scala/com/softwaremill/adopttapir/template/scala/APIDefinitionsView.scala
```diff
@@ -51,4 +58,32 @@ object APIDefinitionsView {
       )
     )
   }
+
+  private object DocumentationEndpoint {
+
+    def prepareDocEndpoints(projectName: String, serverEffect: ServerEffect): PlainLogicWithImports = {
+      val endpoints: List[String] = List("helloEndpoint")
+
+      PlainLogicWithImports(prepareCode(projectName, serverEffect, endpoints), prepareImports(serverEffect))
```
adamw, line 67: maybe rename `PlainLogicWithImports` to simply `Code` which has `imports` & `body` sections?

backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterModel.scala
marcin-jozefowicz: makes sense

backend/src/main/scala/com/softwaremill/adopttapir/template/scala/APIDefinitionsView.scala
```diff
@@ -51,4 +58,32 @@ object APIDefinitionsView {
...
+      }
+      s"""  val docEndpoints: List[${effectStr._1}] = SwaggerInterpreter().fromEndpoints[${effectStr._2}](List(${endpoints.mkString(
+          ","
+        )}), "${projectName}", "1.0.0")""".stripMargin
+    }
+
+    def prepareImports(serverEffect: ServerEffect): List[Import] = {
+      Import("sttp.tapir.swagger.bundle.SwaggerInterpreter") ::
+        (serverEffect match {
+          case ServerEffect.ZIOEffect => List(Import("zio.Task"))
```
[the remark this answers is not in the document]
adamw: Ah ... no, that's fine.

Actually we might want to run scalafix's organize imports on the result (https://github.com/liancheng/scalafix-organize-imports) before zipping. But that's another task :)

backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITSpec.scala
adamw: but looks much bette r:)

adamw: well if you call it `sbtCompileTest(tempDir)` it will be obvious as well and the number of strings will decrease ;)

marcin-jozefowicz: In that particular case when all tests are different only in request I think Your idea will be better. I will hide test method back to helper function.

PR #23, opened 2022-06-08
Adopt-tapir frontend webapp

lubarskyy:


adamw:
I think we should add sth to CI so that the app is at least built? + install the necessary stuff for deploy to work

adamw:
Can you update the readme with what needs to be installed, and how to run locally? :)

lubarskyy:
> I think we should add sth to CI so that the app is at least built? + install the necessary stuff for deploy to work

The `node.js` build environment is included by default on our runner os - https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md, the commands to install and build the webapp are included in the sbt build process. I do believe we should be fine here.

adamw:
some notes:
* `Generate tAPIr starter configuration` -> `Generate tapir project`
* can we validate the project name on the frontend?
* can we have tapir version 1.0.0 pre-selected? there's no other option ;)
* effect type: let's rename `IO` to `cats-effect`
* effect implementation: rename to `server implementation`
* `Documentation endpoint` -> `Expose endpoint documentation using Swagger UI`

adamw:
will the embeddable version come in a subsequent PR? :)

adamw:
> The node.js build environment is included by default on our runner os - https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md, the commands to install and build the webapp are included in the sbt build process. I do believe we should be fine here.

Ah ok, great :) I had to do `npm install -g yarn` locally (after installing node 16 itself) which got me wondering :)

lubarskyy:
> will the embeddable version come in a subsequent PR? :)

The embeddable version 

> some notes:
> 
> * `Generate tAPIr starter configuration` -> `Generate tapir project`
> * can we validate the project name on the frontend?
> * can we have tapir version 1.0.0 pre-selected? there's no other option ;)
> * effect type: let's rename `IO` to `cats-effect`
> * effect implementation: rename to `server implementation`
> * `Documentation endpoint` -> `Expose endpoint documentation using Swagger UI`

all changes applied :+1:

adamw:
> The embeddable version can be already access via iframe element pointing to following path, example below:

Ah great, can you add it to the readme? :)

Also, can we add validation of the group id similar as for project name (except that you can add dots, as in a domain name). Then please merge & we'll see if depolyment works :)

adamw (approved):

PR #26, opened 2022-06-13
For better validation error message corrected regex for projectName p…

marcin-jozefowicz:
…rovided. Will suggest that regex need to match whole input.


lubarskyy (approved):
LGTM :+1:

PR #33, opened 2022-06-14
Adjust template according to requirements

marcin-jozefowicz:
Main: https://gist.github.com/adamw/68ec460dd12878f1218a7d0c9e65f2fe
** println depends on docs 
Endpoints: https://gist.github.com/adamw/d22052eddb1effcb852888b3a87c62e1
** Library as separate object
** section separated by new lines for particular endpoints
** return type for `all`

### DIFF
--- a/backend/src/main/twirl/MainZIOhttpZIO.scala.txt
+++ b/backend/src/main/twirl/MainZIOhttpZIO.scala.txt
@@ -1,16 +1,29 @@
-@(groupId: String)
+@(groupId: String, addDocumentation: Boolean)
 package @groupId
 
 import sttp.tapir.server.ziohttp.ZioHttpInterpreter
 import zhttp.http.HttpApp
-import zhttp.service.Server
+import zhttp.service.server.ServerChannelFactory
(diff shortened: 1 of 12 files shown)


backend/src/main/twirl/MainZIOhttpZIO.scala.txt
```diff
@@ -1,16 +1,29 @@
...
     val app: HttpApp[Any, Throwable] = ZioHttpInterpreter().toHttp(Endpoints.all)
 
-    Server.start(8080, app).exitCode
+    Server(app)
+      .withPort(8080)
+      .make
+      .flatMap(start =>
+        ZIO.succeedBlocking {
+          @if(addDocumentation){println("Go to http://localhost:8080/docs to open SwaggerUI. Press any key to exit.")}else{println("Server started at http://localhost:8080. Press any key to exit.")}
+          StdIn.readLine()
```
adamw, line 23: ZIO has its own way of interacting with the console, see `Console` - which yields `ZIO` effects

PR #54, opened 2022-07-08
feat: add support for metrics endpoint - backend

geminicaprograms:
Add support in backend for generation of the prometheus metrics
endpoint.
The following server implementations are supported:
* Akka
* Http4s
* ZIOHttp

As of now Tapir `1.0.0` has no support for `Netty` server options customisations hence its support cannot be added.

TODO:
- [x]  add frontend implementation

Related to #6.


backend/src/main/scala/com/softwaremill/adopttapir/template/scala/EndpointsView.scala
```diff
@@ -173,12 +182,30 @@ object EndpointsView {
...
+        case ServerEffect.FutureEffect => ("Future", "ServerEndpoint[Any, Future]")
+        case ServerEffect.IOEffect => ("IO", "ServerEndpoint[Any, IO]")
+        case ServerEffect.ZIOEffect => ("Task", "ZServerEndpoint[Any, Any]")
+      }
+      Code(
+        s"""  val prometheusMetrics: PrometheusMetrics[${effect}] = PrometheusMetrics.default[${effect}]()
+           |  val metricsEndpoint: ${endpoint} = prometheusMetrics.metricsEndpoint""".stripMargin,
+        Set(
+          Import("sttp.tapir.server.metrics.prometheus.PrometheusMetrics")
+        )
```
geminicaprograms, line 197: missing `Import("zio.Task")` which is detected when documentation is not requested

geminicaprograms: Done

backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequestValidator.scala
```diff
@@ -125,6 +133,18 @@ sealed trait FormValidator {
     }
   }
 
+  private def validateEffectImplementationWithMetrics(
```
marcin-jozefowicz, line 136: What you think about returning only the validated part? 

```
  private def validateMetricsWithEffectAndImplementation(
      effect: EffectRequest,
      serverImplementation: ServerImplementationRequest,
      addMetrics: Boolean
  ): ValidatedNec[RequestValidation, Boolean] = {
    (effect, serverImplementation, addMetrics) match {
      case t @ (_, Akka | Http4s | ZIOHttp, _) => t._3.validNec
      case t @ (_, Netty, false)               => t._3.validNec
      case (_, Netty, true) => RequestValidation.MetricsNotSupportedForNettyServerImplementation(effect, serverImplementation).invalidNec
    }
  }
```

Then you can reuse result. 
```
  def validate(r: StarterRequest): Either[IncorrectInput, StarterDetails] =
    (
      validateSemanticVersioning(r.tapirVersion),
      validateProjectName(r.projectName),
      validateGroupId(r.groupId),
      validateEffectWithImplementation(r.effect, r.implementation),
      validateMetricsWithEffectAndImplementation(r.effect, r.implementation, r.addMetrics),
      validateEffectWithJson(r.effect, r.json)
    ).mapN { case (tapirVersion, projectName, groupId, (effect, serverImplementation), addMetrics, json) =>
      StarterDetails(
        projectName,
        groupId,
        effect.toModel,
        serverImplementation.toModel,
        tapirVersion,
        r.addDocumentation,
        addMetrics,
        json.toModel
      )
    }.leftMap(errors => IncorrectInput(errors.toNonEmptyList.map(_.errMessage).toList.mkString(System.lineSeparator())))
      .toEither
```

backend/src/main/scala/com/softwaremill/adopttapir/template/scala/EndpointsView.scala
```diff
@@ -173,12 +182,28 @@ object EndpointsView {
 
   }
 
+  private object MetricsEndpoint {
+    def prepareMetricsEndpoint(serverEffect: ServerEffect): Code = {
+      val (effect, endpoint) = serverEffect match {
```
marcin-jozefowicz, line 187: There is redundancy with doc endpoints.  EndpointsView:212.  Maybe you could extract method for this common code.

backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITTest.scala
```diff
@@ -17,11 +17,13 @@ class StarterServiceITTest extends BaseTest with ParallelTestExecution {
     effect <- EffectRequest.values
```
marcin-jozefowicz, line 17: What do you think about creating an IT test  where we run: 

"sbt run " in one thread and call /metrics from another? It will then check if metrics are really available under endpoint.

backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequestValidator.scala
geminicaprograms: Good point! Was wondering how to achieve it ;)

backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITTest.scala
geminicaprograms: I like the idea but see the following issues with that:
* we would need some means to wait until the endpoint is ready and stop it once the verification is done
* it would further extend the tests run (and it will increase exponentially with number of options) - I can imagine it is done as the next (conditional) step of [StarterServiceITTest.scala#L44](https://github.com/softwaremill/adopt-tapir/blob/main/backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITTest.scala#L44)

Should we add it as a follow up? WDYT?

marcin-jozefowicz: According to documentation: 
https://www.lihaoyi.com/post/HowtoworkwithSubprocessesinScala.html#osprocspawn there is possibility to spawn 2 processess and feed one from other's output. So technically there is no need to introduce wait. Just getting that service is up and running. (We have welcome input for all implementation I guess) 

The another approach will be to add separate test in generated zip project which will run server and check if metrics are available.  Then Integration tests will not be touched. 

It's up to Your preference to split this into separate issue or do it in this task. I am good with both directions.

backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequestValidator.scala
```diff
@@ -72,15 +78,17 @@ sealed trait FormValidator {
       validateProjectName(r.projectName),
       validateGroupId(r.groupId),
       validateEffectWithImplementation(r.effect, r.implementation),
+      validateEffectImplementationWithMetrics(r.effect, r.implementation, r.addMetrics),
```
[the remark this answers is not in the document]
geminicaprograms: doh, I have missed this one :/ will address it promptly once the chain gets merged ;)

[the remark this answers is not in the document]
geminicaprograms: Done as I have messed up with merges - wanted to merge this one prior the other one... ;)

marcin-jozefowicz (approved):

PR #55, opened 2022-07-11
feat: enable 'addMetrics' config option in UI by default

geminicaprograms:
Changes:
* remove feature flag `REACT_APP_METRICS_ENDPOINTS_FF` and related
  functionality
* metrics are not supported when server implementation is `Netty`
  therefore add logic that disables `addMetrics` configuration in case
  when server implementation is not specified or points to `Netty`

Related to #6.


ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
```diff
@@ -64,6 +66,11 @@ export const ConfigurationForm: React.FC<ConfigurationFormProps> = ({ isEmbedded
     ) {
       form.resetField('json');
     }
+
+    // NOTE: reset addMetrics field value if metrics are not supported
+    if (!isAddMetricsSupported(effectImplementation)) {
```
lubarskyy, line 71: The form typing is not great, but maybe we could move the `Boolean(effectImplementation)` check from `isAddMetricsSupported` function to this if statement? It would then follow the pattern of previous cases:

```typescript
if (effectImplementation && !isAddMetricsSupported(effectImplementation)) {
   form.resetField('addMetrics');
}
```

```typescript
export const isAddMetricsSupported = (effectImplementation: EffectImplementation): boolean => {
  return effectImplementation !== EffectImplementation.Netty;
}
```
let me know what you think :)

lubarskyy (commented):
Please run `yarn lint` in `ui` directory after you are finished with your changes, this command will both check for linting errors and properly format code.

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
geminicaprograms: We could do that but condition IMHO should be
```
if (!effectImplementation || !isAddMetricsSupported(effectImplementation)) {
   form.resetField('addMetrics');
}
```
so that field gets reset when `effectImplementaiton` has no value or when it equals to `EffectImplementation.Netty`, WDYt?

lubarskyy: we want to reset this form field when `effectImplementation` is defined and it's value is equal to `EffectImplementation.Netty`, if we make an `OR` here with `effectImplementation` being an undefined value, we would be resetting `addMetrics` form field every time we update one of the dependencies of this `useEffect` hook.

Nevertheless, we are also using this `isAddMetricsSupported` function down in the form group component and if we remove the boolean check from it, it breaks the initial disabled state. So I would say let's keep your original implementation :) :+1:

lubarskyy (approved):
LGTM :+1:

PR #77, opened 2022-07-15
chore: add docs endpoint verification to framework endpoints spec

geminicaprograms:
When `addDocumentation` is selected and `StarterServiceITTest` is called
generate `FrameworkEndpointsSpec.scala` that verifies if docs
endpoint is available and returns expected result.
Note that the spec in question is only compiled for tests and doesn't
polute the generated production code.

For development purposes one can examine the spec by calling:
```
  com.softwaremill.adopttapir.starter.FileOperation
```
end examining their output.

### DIFF
--- a/backend/src/test/twirl/FrameworkEndpointsSpecZIO.scala.txt
+++ b/backend/src/test/twirl/FrameworkEndpointsSpecZIO.scala.txt
@@ -1,7 +1,8 @@
 @(
 starterDetails: com.softwaremill.adopttapir.starter.StarterDetails,
 additionalImports: List[com.softwaremill.adopttapir.template.scala.Import],
-metricsStub: String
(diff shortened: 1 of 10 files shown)


marcin-jozefowicz (approved):

backend/src/test/twirl/FrameworkEndpointsSpecZIO.scala.txt
```diff
@@ -38,6 +39,21 @@ object FrameworkEndpointsSpec extends ZIOSpecDefault {
         )
       )
     }
+}@if(metricsStub.nonEmpty){@if(metricsStub.nonEmpty){,}
```
geminicaprograms, line 42: in the first condition it should be `docsStub` ;)

PR #89, opened 2022-07-28
Feat/test yday

geminicaprograms:


geminicaprograms:
it was done for testing only

PR #93, opened 2022-07-29
Scala3 support rebased

marcin-jozefowicz:
PR is about extending templates to support also Scala 3. 

## Notes: 
- `Akka` implementation is excluded from supporting scala 3, 
- Versions of scala 2 & scala 3 are taken from `build.sbt` to allow updating it automatically by `scala-steward`
- Used materials: 
  - https://docs.scala-lang.org/scala3/guides/migration/tooling-syntax-rewriting.html
 

An example of UI after changes: 
Screenshot 2022-08-01 at 14 23 09
image


ui/src/components/ConfigurationForm/ConfigurationForm.consts.ts
```diff
@@ -124,28 +128,28 @@ export const createStarterValidationSchema = (isScalaVersionFieldVisible: boolea
...
       implementation: yup
         .mixed()
-        .when('effect', {
+        .when(['effect'], {
           is: EffectType.Future,
           then: schema =>
             schema.oneOf(
               mapEffectTypeToEffectImplementation(EffectType.Future),
-              getEffectImplementationFieldMessage(EffectType.Future)
+              getEffectImplementationFieldMessage(EffectType.Future, ScalaVersion.Scala2)
```
marcin-jozefowicz, line 136: @lubarskyy  - I am not sure if this is properly handled. Can You look on UI changes?

geminicaprograms:
I was wondering about rationale behind drop-down list for scala versions as simple toggle button (with `Scala 2` preselected) would be easier to use (requires at most 1 click). Do we plan to support more versions?

lubarskyy:
> I was wondering about rationale behind drop-down list for scala versions as simple toggle button (with `Scala 2` preselected) would be easier to use (requires at most 1 click). Do we plan to support more versions?

Previously we had a tapir version select, it could potentially consist of multiple entries due to semver, naturally scala version followed same pattern due to visual similarities. If we are sure that tapir version selector is not coming back I think we could switch this to radio buttons :+1:

marcin-jozefowicz:
@geminicaprograms 
>I was wondering about rationale behind drop-down list for scala versions as simple toggle button (with Scala 2 preselected) would be easier to use (requires at most 1 click). Do we plan to support more versions?
  
UI part for scala version was already created in such form and accepted previously. If you feel that toggle button will be better I could change it.

adamw:
There's going to be no tapir version selection, and Scala will only have two options: 2 and 3.

ui/src/components/ConfigurationForm/ConfigurationForm.consts.ts
```diff
@@ -87,15 +91,15 @@ const valueGetter = (
...
+const getEffectImplementationFieldMessage = (effectType: EffectType, scalaVer: ScalaVersion): string =>
+  `Server implementation must be one of the following values: ${getEffectImplementationOptions(effectType, scalaVer)
     .map(labelGetter)
     .join(', ')}`;
 
 const REQUIRED_FIELD_MESSAGE = 'This field is required';
 
-export const createStarterValidationSchema = (isScalaVersionFieldVisible: boolean) => {
-  const baseSchema = yup
+export const createStarterValidationSchema = () => {
```
lubarskyy, line 101: This no longer takes parameters, we can safely turn it into a static object.

```typescript
const starterValidationSchema = {
  ...
};
```

ui/src/components/ConfigurationForm/ConfigurationForm.consts.ts
```diff
@@ -124,28 +128,28 @@ export const createStarterValidationSchema = (isScalaVersionFieldVisible: boolea
       // NOTE: unfortunately this is the only way of multiple .when cases in yup :shrug-emoji:
       implementation: yup
         .mixed()
-        .when('effect', {
+        .when(['effect'], {
```
lubarskyy, line 131: leftover array notation?

geminicaprograms:
> UI part for scala version was already created in such form and accepted previously. If you feel that toggle button will be better I could change it.

We can also address it in the follow-up if it would make things easier ;)

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
```diff
@@ -155,7 +154,7 @@ export const ConfigurationForm: React.FC<ConfigurationFormProps> = ({ isEmbedded
             name="implementation"
             label="Server implementation"
             disabled={!isEffectTypeSelected}
-            options={isEffectTypeSelected ? getEffectImplementationOptions(effectType) : []}
+            options={isEffectTypeSelected ? getEffectImplementationOptions(effectType, scalaVer) : []}
```
lubarskyy, line 157: Ok, so in my opinion here is how server implementation select should behave. Currently we disable this field if user did not select any effect type, with the recent change of adding a scala version, effect implementation has become also dependent on that new value.

so const `isEffectTypeSelected` should be renamed to something like:
```typescript
const isEffectImplementationSelectable = Boolean(effectType) && Boolean(scalaVer);

...

<FormSelect
  className={classes.formEffectsRow}
  name="implementation"
  label="Server implementation"
  disabled={!isEffectImplementationSelectable}
  options={isEffectImplementationSelectable ? getEffectImplementationOptions(effectType, scalaVer) : []}
/>
```
we should disable the select until user selects both effect type and scala version, we want to mitigate scenarios like below, for `Future` effect type user cannot select `Akka` because he hasn't selected a scala version. For someone who is here first time this is a discoverability problem.
2022-08-01-151634_1031x654_scrot

lubarskyy: Additionally we have to update the `useEffect` hook in order to reset server implementation field upon scala version change:
```typescript
useEffect(() => {
  ...
  // NOTE: reset effect implementation field value upon scala version change
  if (
    scalaVer &&
    scalaVer === ScalaVersion.Scala3 &&
    effectImplementation &&
    forbiddenScala3EffectImplementations.includes(effectImplementation)
  ) {
    form.resetField('implementation');
  }
  ...
}, [effectType, scalaVer, effectImplementation, jsonImplementation, form]);
```

ui/src/components/ConfigurationForm/ConfigurationForm.helpers.ts
```diff
@@ -17,10 +17,13 @@ export const mapEffectTypeToEffectImplementation = (effectType: EffectType): Eff
   return effectTypeImplementationMap[effectType];
 };
 
-export const getEffectImplementationOptions = (effectType: EffectType): FormSelectOption[] => {
+export const getEffectImplementationOptions = (effectType: EffectType, scalaVer: ScalaVersion): FormSelectOption[] => {
```
lubarskyy, line 20: Maybe a bit more verbose? Ideally this should be piped, I need to add some library to help with this.

```typescript
export const getEffectImplementationOptions = (effectType: EffectType, scalaVer: ScalaVersion): FormSelectOption[] => {
  const mappedEffectImplementations = mapEffectTypeToEffectImplementation(effectType);
  const forbiddenScala3EffectImplementations = [EffectImplementation.Akka];

  const availableEffectImplementations = mappedEffectImplementations.filter(effectImplementation =>
    scalaVer === ScalaVersion.Scala2
      ? effectImplementation
      : !forbiddenScala3EffectImplementations.includes(effectImplementation)
  );

  return EFFECT_IMPLEMENTATIONS_OPTIONS.filter(({ value }) => availableEffectImplementations.includes(value));
};
```

marcin-jozefowicz:
UI after changes: 
image

build.sbt
```diff
@@ -230,12 +231,14 @@ lazy val backend: Project = (project in file("backend"))
   .enablePlugins(SbtTwirl)
   .settings(dockerSettings)
 
+val scala3Version = "3.1.1"
```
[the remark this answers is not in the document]
marcin-jozefowicz: done

backend/src/main/twirl/Endpoints.scala.txt
```diff
@@ -6,29 +6,33 @@ metricsEndpoint: String,
...
 import sttp.tapir.{PublicEndpoint, endpoint, query, stringBody}
 @for(additionalImport <- additionalImports) {
-@{additionalImport.asScalaImport()}}
+@{additionalImport.asScalaImport(scalaVersion)}}
 
-object Endpoints {
-  case class User(name: String) extends AnyVal
-  val helloEndpoint: PublicEndpoint[User, Unit, String, Any] = endpoint.get
+object Endpoints@if(scalaVersion == com.softwaremill.adopttapir.starter.ScalaVersion.Scala2){ @leftBracket}else{:}
+@{indent}case class User(name: String) extends AnyVal
```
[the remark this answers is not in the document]
marcin-jozefowicz: Moved to `EndpointsView`

adamw:
@marcin-jozefowicz looks good, though we can probably make it nicer (UI/UX-wise), maybe @lubarskyy can suggest some layout :)

backend/src/main/twirl/Endpoints.scala.txt
```diff
@@ -6,29 +6,33 @@ metricsEndpoint: String,
...
+@{indent}val helloEndpoint: PublicEndpoint[User, Unit, String, Any] = endpoint.get
     .in("hello")
     .in(query[User]("name"))
     .out(stringBody)
-@helloEndpointServer
+@{indent}@helloEndpointServer
 
-@jsonEndpoint
+@{indent}@jsonEndpoint
```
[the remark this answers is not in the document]
marcin-jozefowicz: Moved to EndpointsView

ui/.env
```diff
@@ -1 +1 @@
-REACT_APP_SCALA_VERSION_FF=false
```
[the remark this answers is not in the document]
marcin-jozefowicz: done

ui/src/components/ConfigurationForm/ConfigurationForm.consts.ts
marcin-jozefowicz: done

ui/src/components/ConfigurationForm/ConfigurationForm.consts.ts
marcin-jozefowicz: Indeed, removed.

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
marcin-jozefowicz: Thx! I haven't noticed reset functionality!  Done.

ui/src/components/ConfigurationForm/ConfigurationForm.helpers.ts
marcin-jozefowicz: Done

lubarskyy:
> @marcin-jozefowicz looks good, though we can probably make it nicer (UI/UX-wise), maybe @lubarskyy can suggest some layout :)

@adamw, @marcin-jozefowicz I would merge it as it is, I need sometime to think about potential solution here -> let's create an issue for that.

adamw:
@lubarskyy yeah sure, as a follow-up

lubarskyy (approved):
The UI part LGTM :+1:

PR #104, opened 2022-08-03
fix(): rework effect implementation field validation, adjust error message

lubarskyy:
- split reset field logic into multiple useEffects,
- rework effect implementation field validation to be based on effect type and scala version,
- add react-hook-form devtools and enable them in development environment for ease of debugging.


ui/src/components/ConfigurationForm/ConfigurationForm.helpers.ts
```diff
@@ -1,10 +1,11 @@
+import difference from 'lodash.difference';
```
marcin-jozefowicz, line 1: Nice ! 
It's now easier to read

marcin-jozefowicz (approved):

lubarskyy:
2022-08-05-152928_973x598_scrot

marcin-jozefowicz:
Looks better!

PR #115, opened 2022-08-09
test(): adopt tapir UI test suites

lubarskyy:


.github/workflows/adopt-tapir-ci.yml
```diff
@@ -60,6 +60,20 @@ jobs:
...
+        run: yarn build
+        working-directory: ./ui
+
+      - name: Run UI tests
+        # run this step only once across all configurations
+        if: matrix.scala == 'Scala3' && matrix.json == 'No'
+        id: run-ui-tests
+        run: yarn test
+        working-directory: ./ui
+
```
geminicaprograms, line 76: IMHO it should be either UI build & test here or added to root project in `build.sbt`.
I would prefer the former as we could keep it separated and called when needed IOW it wouldn't impact backend development until one needs it (consider adding that part to `README.md` if it is not there yet) as now it is called two times (both compilation and unit tests).
Here is the [job log](https://github.com/softwaremill/adopt-tapir/runs/7777186105):
1. compilation is part of the `Run unit tests` step
2. UI tests are part of the `Run unit tests` step
3. they are repeated when these two steps are called

geminicaprograms:
There is this warning in UI tests (printed in [Run UI tests](https://github.com/softwaremill/adopt-tapir/runs/7777186105?check_suite_focus=true))
```
console.error
      The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to ":first-of-type".
```
Is this is sth that could/should be fixed?

geminicaprograms:
Calling `sbt test` never finishes locally:
image

.github/workflows/adopt-tapir-ci.yml
```diff
@@ -60,6 +60,20 @@ jobs:
         id: build-docker-image-locally
         run: sbt docker:publishLocal
 
+      - name: Run UI build
+        # run this step only once across all configurations
+        if: matrix.scala == 'Scala3' && matrix.json == 'No'
+        id: run-ui-build
+        run: yarn build
+        working-directory: ./ui
```
marcin-jozefowicz, line 68: Just put  in package.json
"test": "react-scripts test a --watchAll=false",

The default process use watch and never ends

lubarskyy:
> There is this warning in UI tests (printed in [Run UI tests](https://github.com/softwaremill/adopt-tapir/runs/7777186105?check_suite_focus=true))
> 
> ```
> console.error
>       The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to ":first-of-type".
> ```
> 
> Is this is sth that could/should be fixed?

Nope, we are not doing a SSR right now so it doesn't concern us :)

marcin-jozefowicz (approved):

Issue #45, opened 2022-06-22
Consider adding scala-cli as an alternative build tool


adamw:
Could be a good fit a small, simple project?

https://scala-cli.virtuslab.org

PR #119, opened 2022-08-09
feat: add scala-cli builder as backend generation option

geminicaprograms:
Add `scala-cli` build tool as `sbt` alternative for generated tapir code in backend.

The following changes were applied:
* new `scala-cli` _builder_ (`sbt` builder value was added too) option was introduced to request
* `scala-cli` project contains (apart from sources) only relevant _project_ files (`build.sc` for project definition and production dependencies, `src/test/scala/test.sc` for test dependencies and `.scalafmt.conf` for formatter)
* dedicated `README.md` was added

Note that additional configuration is covered in integration tests which were also refactored (Service and Describers landed in a dedicated files).

Part of #45


backend/src/main/scala/com/softwaremill/adopttapir/starter/FilesGenerator.scala
```diff
@@ -0,0 +1,53 @@
...
+}
+
+private case object ScalaCliFilesGenerator extends FilesGenerator {
+  private lazy val scalaCliTemplate: ScalaCliProjectTemplate = new ScalaCliProjectTemplate()
+
+  override def generate(starterDetails: StarterDetails): List[GeneratedFile] = {
+    List(
+      scalaCliTemplate.getBuildScalaCli(starterDetails),
+      scalaCliTemplate.getTestScalaCli(starterDetails),
+      scalaCliTemplate.getMain(legalizeGroupId(starterDetails)),
```
marcin-jozefowicz, line 46: Method SbtProjectTemplate#legalizeGroupId should be moved to ProjectTemplate as it is used by both generators. 

Maybe make it private and run under body of .getMain / .getEndpoints / .getEndpointsSpec .

backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequest.scala
```diff
@@ -11,7 +11,8 @@ case class StarterRequest(
     addDocumentation: Boolean,
     addMetrics: Boolean,
     json: JsonImplementationRequest,
-    scalaVersion: ScalaVersionRequest
+    scalaVersion: ScalaVersionRequest,
+    builder: BuilderRequest = BuilderRequest.Sbt
```
marcin-jozefowicz, line 15: I am aware that this is needed for backward compatibility. 

Please add at least TODO: to not forget to remove it during providing UI changes.

geminicaprograms: Added

backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITTest.scala
```diff
@@ -247,14 +248,65 @@ case class Service(tempDir: better.files.File) {
   }
 }
 
+case class SbtService(tempDir: better.files.File) extends Service {
```
marcin-jozefowicz, line 251: As it is used only by SbtService object it can be hidden there: 

```
object SbtService {
  def apply(tempDir: better.files.File): IO[SbtService] = {
    IO.blocking(SbtService(tempDir))
  }

  private case class SbtService(tempDir: better.files.File) extends Service {
    override protected val portPattern = new Regex("^(?:\\[info\\] )(?:Go to |Server started at )http://localhost:(\\d+).*")

    override protected lazy val process: SubProcess = {
      os.proc(
        "sbt",
        "-no-colors",
        // start in forked mode so that process input can be forwarded and process waits before closing otherwise `StdIn.readLine` will exit immediately
        "set run / fork := true",
        // forward std input to forked process - https://www.scala-sbt.org/1.x/docs/Forking.html#Configuring+Input
        "set run / connectInput := true",
        ";compile ;test ;run"
      ).spawn(cwd = os.Path(tempDir.toJava), env = Map("http.port" -> "0"), mergeErrIntoOut = true)
    }
  }
```

backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITTest.scala
```diff
@@ -6,7 +6,6 @@ import cats.effect.unsafe.implicits.global
 import cats.instances.list._
```
marcin-jozefowicz, line 6: I think that splitting the file into smaller one could increase the readability. 

WDYT?

backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITTest.scala
marcin-jozefowicz: Same is true for Scala Cli

marcin-jozefowicz:
Update `./development/http-client.http` with new enum

backend/src/main/scala/com/softwaremill/adopttapir/starter/FilesGenerator.scala
geminicaprograms: good point, done!

backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITTest.scala
marcin-jozefowicz: Or even

object Service {
    def sbt (starterDetails: StarterDetails): IO[Service] = ??? 
    def scalaCLI(starterDetails: StarterDetails): IO[Service] = ??? 
}

so you will have one object responsible to provide different implementations

backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITTest.scala
```diff
@@ -49,27 +49,43 @@ object Setup {
...
+object Describers {
+  implicit class StarterDetailsWithDescribe(details: StarterDetails) {
+    lazy val describe: String =
+      s"${details.serverEffect}/${details.serverImplementation}/docs=${details.addDocumentation}/metrics=${details.addMetrics}" +
+        s"/json=${details.jsonImplementation}/scalaVersion=${details.scalaVersion}/builder=${details.builder}"
+  }
+
+  implicit class ThrowableWithDescribe(e: Throwable) {
+    def describe(): String = {
+      Using.Manager { _ =>
```
marcin-jozefowicz, line 77: According to api doc: 

> Performs an operation using a Manager, then closes the Manager, releasing its resources (in reverse order of acquisition).
Example:
 val lines = Using.Manager { use =>
   use(new BufferedReader(new FileReader("file.txt"))).lines()
 }
 
> If using resources which require an implicit Manager as a parameter, this method should be invoked with an implicit modifier before the function parameter:
Example:
 val lines = Using.Manager { implicit use =>
   new SafeFileReader("file.txt").lines()
 }
   
 Looks like resources are not cleared after execution of method. 
 
 I would expect something like: 
```
    def describe(): String = {
      Using.Manager { use =>
        val sw = use(new StringWriter())
        val pw = use(new PrintWriter(sw))
        e.printStackTrace(pw)
        sw.toString
      }.get
    }
```

backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITTest.scala
geminicaprograms: I have introduced object `Service` that hides selection of implementation depending on a builder type
```
object Service {
  import Timeouts.waitForScalaCliCompileAndUnitTest

  def apply(builder: Builder, tempDir: better.files.File): IO[Service] = {
    builder match {
      case Builder.Sbt      => IO.blocking(SbtService(tempDir))
      case Builder.ScalaCli => IO.blocking(ScalaCliService(tempDir)).timeoutAndForget(waitForScalaCliCompileAndUnitTest)
    }
  }

  private case class SbtService(tempDir: better.files.File) extends Service {
  ...
  }

  private case class ScalaCliService(tempDir: better.files.File) extends Service {
  ...
  }
```

backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITTest.scala
geminicaprograms: I have somehow misinterpreted it :/ thanks for catching it up!

backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITTest.scala
```diff
@@ -6,20 +6,14 @@ import cats.effect.unsafe.implicits.global
...
-import scala.annotation.tailrec
 import scala.collection.mutable
-import scala.concurrent.TimeoutException
 import scala.concurrent.duration.{DurationInt, FiniteDuration}
-import scala.util.{Properties, Using}
-import scala.util.matching.Regex
+import scala.util.Properties
 
 object Setup {
   lazy val jsonImplementations: List[JsonImplementationRequest] = {
```
marcin-jozefowicz, line 19: I propose to hide the variables which are not use externally. It helps in readability what is used outside. 

```
object Setup {
  type TestFunction = Integer => IO[Unit]

  val validConfigurations: Seq[StarterDetails] = for {
    effect <- EffectRequest.values
    server <- ServerImplementationRequest.values
    docs <- List(true, false)
    metrics <- List(true, false)
    json <- jsonImplementations
    scalaVersion <- scalaVersions
    builder <- BuilderRequest.values
    starterRequest = StarterRequest(
      "myproject",
      "com.softwaremill",
      effect,
      server,
      addDocumentation = docs,
      addMetrics = metrics,
      json,
      scalaVersion,
      builder
    )
    starterDetails <- FormValidator.validate(starterRequest).toSeq
  } yield starterDetails

  private lazy val jsonImplementations: List[JsonImplementationRequest] = {
    Properties
      .envOrElse("JSON", JsonImplementationRequest.values.mkString(","))
      .split(",")
      .map(JsonImplementationRequest.withName)
      .toList
  }

  private lazy val scalaVersions: List[ScalaVersionRequest] = {
    Properties.envOrElse("SCALA", ScalaVersionRequest.values.mkString(",")).split(",").map(ScalaVersionRequest.withName).toList
  }
}
```

backend/src/test/scala/com/softwaremill/adopttapir/test/Describers.scala
```diff
@@ -0,0 +1,25 @@
+package com.softwaremill.adopttapir.test
+
+import com.softwaremill.adopttapir.starter.StarterDetails
+
+import java.io.{PrintWriter, StringWriter}
+import scala.util.Using
+
+object Describers {
```
marcin-jozefowicz, line 8: JFYI - not required any changes. 

This approach is similar to TypeClasses. 

```
trait Describer[A] {
  def describe(value: A): String
}

object DescriberInstances {
  implicit val detailsDescriber = new Describer[StarterDetails] {
    override def describe(details: StarterDetails): String =
      s"${details.serverEffect}/${details.serverImplementation}/docs=${details.addDocumentation}/metrics=${details.addMetrics}" +
        s"/json=${details.jsonImplementation}/scalaVersion=${details.scalaVersion}/builder=${details.builder}"
  }

  implicit val throwableDescriber = new Describer[Throwable] {
    override def describe(e: Throwable): String = {
      Using.Manager { use =>
        val sw = use(new StringWriter())
        val pw = use(new PrintWriter(sw))
        e.printStackTrace(pw)
        sw.toString
      }.get
    }
  }
}

object DescriberSyntax {
  implicit class DescriberOps[A](value: A) {
    def describe(implicit d: Describer[A]): String = d.describe(value)
  }
}
```

And as it is similar to cats Show https://typelevel.org/cats/typeclasses/show.html 

You can just use build-in class type and only provide expected instances. The rest - syntax will be for free. 

```
object ShowInstances {
  implicit val dShow = Show.show[StarterDetails](details =>
    s"${details.serverEffect}/${details.serverImplementation}/docs=${details.addDocumentation}/metrics=${details.addMetrics}" +
      s"/json=${details.jsonImplementation}/scalaVersion=${details.scalaVersion}/builder=${details.builder}"
  )

  implicit val throwableShow = Show.show[Throwable] { e =>
    Using.Manager { use =>
      val sw = use(new StringWriter())
      val pw = use(new PrintWriter(sw))
      e.printStackTrace(pw)
      sw.toString
    }.get

  }
}

```
and then use show type class whenever you want

marcin-jozefowicz: Long story short: You don't have to create another abstraction - cats have build-in solution for U.

marcin-jozefowicz:
This is general comment. 

For me as a reader it is hard to follow all those global singleton objects. It is hard to reason about from where changes comes.  Right now it is manageable. But number of exposed elements takes time to follow or check.

geminicaprograms:
> This is general comment.
> 
> For me as a reader it is hard to follow all those global singleton objects. It is hard to reason about from where changes comes. Right now it is manageable. But number of exposed elements takes time to follow or check.

Reduced the dependency on global object by either injection (production code) or constructor parameters (IT tests). `FilesGenerator` functionality was combined into `ProjectTemplate` (now called `ProjectGenerator`).

marcin-jozefowicz:
For me looks good :)

marcin-jozefowicz (approved):

backend/src/main/scala/com/softwaremill/adopttapir/template/ProjectGenerator.scala
```diff
@@ -0,0 +1,212 @@
...
+}
+
+object CommonObjectTemplate {
+  def templateResource(fileName: String): String = Resource.getAsString(s"template/$fileName")
+
+  val scalafmtConfigPath = ".scalafmt.conf"
+
+  val readMePath: String = "README.md"
+
+  implicit class StarterDetailsWithLegalizedGroupId(starterDetails: StarterDetails) {
```
adamw, line 168: maybe we should simply change the requirements for the package name so that invalid group ids are rejected?

backend/src/main/scala/com/softwaremill/adopttapir/template/ProjectGenerator.scala
```diff
@@ -0,0 +1,212 @@
...
+      }
+    }
+
+    def legalizeGroupId(groupId: String): String = groupId.split('.').map(legalize).mkString(".")
+
+    legalizeGroupId(starterDetails.groupId)
+  }
+}
+
+private object ScalaCliProjectTemplate extends ProjectTemplate {
```
adamw, line 189: [minor] I think instead of overriding methods I'd define methods such as `generateCode` and `generateBuild`, and then the `generate` impl would be `generateCode ++ generateBuild`

PR #123, opened 2022-08-11
BUGFIX - load classes

marcin-jozefowicz:


marcin-jozefowicz:
Picked another approach.

PR #136, opened 2022-08-18
Improvements to library

rafalambrozewicz:
* Removed redundant `AtomicReference` wrapping list of books.
* Changed the way `books` are referenced.
* Updated integration tests

### DIFF
--- a/backend/src/main/scala/com/softwaremill/adopttapir/template/scala/EndpointsView.scala
+++ b/backend/src/main/scala/com/softwaremill/adopttapir/template/scala/EndpointsView.scala
@@ -122,10 +122,7 @@ object EndpointsView {
          |    Book("Pharaoh", 1897, Author("Boleslaw Prus"))
          |  )
          |)
-         |${if (scalaVersion == Scala2) "}" else ""}""".stripMargin,
-      Set(
-        Import("Library._")
-      )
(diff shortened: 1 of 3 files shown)


backend/src/main/scala/com/softwaremill/adopttapir/template/scala/EndpointsView.scala
```diff
@@ -115,17 +115,14 @@ object EndpointsView {
...
+         |  val books = List(
+         |    Book("The Sorrows of Young Werther", 1774, Author("Johann Wolfgang von Goethe")),
+         |    Book("On the Niemen", 1888, Author("Eliza Orzeszkowa")),
+         |    Book("The Art of Computer Programming", 1968, Author("Donald Knuth")),
+         |    Book("Pharaoh", 1897, Author("Boleslaw Prus"))
+         |  )
          |${if (scalaVersion == Scala2) "}" else ""}""".stripMargin,
       Set(
-        Import("java.util.concurrent.atomic.AtomicReference"),
         Import("Library._")
```
adamw, line 126: do we still need that import?

rafalambrozewicz: Yes, since `Library.Book` and `Library.Author` is referenced in generated code, without it, the code won't compile.

PR #142, opened 2022-08-23
Creating list of endpoints once

rafalambrozewicz:
* In generated project list of API endpoints is created once
* Updated endpoints order in case classes and template
* In `EndpointsView.scala` updated code order, so that it corresponds to the aforementioned changes


adamw:
Thanks! :)

PR #143, opened 2022-08-23
Make backend fronted work locally with each other

rafalambrozewicz:
* On frontend side - allowed changing server address by specifying environmental variable, so that it is possible to easily plug it to localhost backend.
* Updated `README.md` accordingly.


adamw:
Thanks!

PR #144, opened 2022-08-23
removed parent object from template, as @main annotated function does…

### LINKED ISSUE #132: When generating a Scala 3 project, main doesn't have to be wrapped in a Main object
It's enough to have a top-level `@main def run(): Unit`

tpacuszkapl:
closes #132


tpacuszkapl:
Changed only `MainFutureNettyScala3.scala.txt` as other templates with scala3 don't have `@main` (they override parent methods).

adamw:
Thanks!

PR #150, opened 2022-08-25
Enable metrics interceptor for netty

rafalambrozewicz:
* Updated project templates for Future&Netty and IO&Netty cases, so that metrics could be added to them.
* Updated `MainView` cases to support adding metrics for Netty projects, adjusted metrics validation in `StarterRequestValidator`.
* Updated frontend to accommodate metrics for Netty cases (removal of no longer needed validation)


backend/src/main/twirl/MainFutureNettyScala3.scala.txt
```diff
@@ -1,18 +1,22 @@
...
 
 import scala.concurrent.duration.Duration
 import scala.concurrent.{Await, ExecutionContext, Future}
 import scala.io.StdIn
 import ExecutionContext.Implicits.global
 
 @@main def run(): Unit =
+  @if(addMetrics) {
+  val serverOptions = NettyFutureServerOptions.default
+    .appendInterceptor(Endpoints.prometheusMetrics.metricsInterceptor())
```
adamw, line 14: `CustomiseInterceptors` has a `.metricsInterceptor` method which adds the interceptor in the correct order (should be prepended, not appended)

rafalambrozewicz: Thanks 👍  updated for all cases.

backend/src/main/twirl/MainFutureNetty.scala.txt
```diff
@@ -10,9 +10,13 @@ import ExecutionContext.Implicits.global
 
 object Main {
   def main(args: Array[String]): Unit = {
+    @if(addMetrics) {
+      val serverOptions = NettyFutureServerOptions.default
+           .prependInterceptor(Endpoints.prometheusMetrics.metricsInterceptor())
```
adamw, line 15: why not the dedicated `.metricsInterceptor`? ;)

adamw:
Thanks :)

PR #156, opened 2022-08-30
View project contents without downloading the project - frontend

tpacuszkapl:
Preview is complete but it still uses the stubbed data for rendering as endpoint is not yet ready. Call to dedicated endpoint will have to be added later.
Change incorporates additional button on the main page (at `/`) and separate page for preview (at `/preview-starter`).
main
preview


tpacuszkapl:
Let me know if you need a demo.

lubarskyy:
@tpacuszkapl General remarks:

### Code formatting
please run `yarn lint` in order to format code according to the specified project settings

### Directory `pages`
each `page` component should have it's own styles definition in a file next to it, project's current convention is as follows:

```
    .
    ├── pages
    │   ├── RootPage
    │       ├── RootPage.component.tsx
    │       ├── RootPage.styles.ts
```
this creates an easy to reason about component structure with clear indication that those styles are only applied to that component, please also note the naming convention of components, we denote it using a `xyz.component.tsx` pattern, it doesn't matter if it's higher in the tree hierarchy, it's still a component :) 

### Solution
Current solution with passing a form state between locations via the History API is very handy, but comes with some drawbacks. First, you need to pass it from location A -> B and then from location B -> A in order to keep the form in sync, having it stored in a `store` of some sorts could ease things up. Second, you need to clear it when user clicks the reset button in the `ConfigurationForm` component when we navigate back to it. One side note is that the location state persist between page refreshes and this might be both annoying and misguiding as this is not standard behavior for forms.

A simple combination of hook `useReducer` with react `Context` - [example implementation](https://medium.com/@mohsentaleb/redux-vs-context-api-usereducer-when-to-use-which-b625b436259) could mimic the redux like behavior. Any component in context of this state would have immediate access to form data.

tpacuszkapl:
@lubarskyy thanks for the review. I will apply suggested changes and let you know when it's ready.

tpacuszkapl:
@lubarskyy I've applied the changes you suggested. Please check if that's what you had in mind.

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
```diff
@@ -57,20 +58,18 @@ export const ConfigurationForm: React.FC<ConfigurationFormProps> = ({ isEmbedded
...
-      form.setValue('groupId', request.groupId);
-      form.setValue('effect', request.effect);
-      form.setValue('implementation', request.implementation);
-      form.setValue('addDocumentation', request.addDocumentation);
-      form.setValue('addMetrics', request.addMetrics);
-      form.setValue('json', request.json);
-      form.setValue('scalaVersion', request.scalaVersion);
-      form.setValue('builder', request.builder);
+    if (formData !== undefined) {
+      form.setValue('projectName', formData.projectName);
```
lubarskyy, line 62: Could we simplify this a bit?
```
Object.entries(formData).map(([key, value]) => form.setValue(key, value));
```

lubarskyy (approved):
LGTM, just one more small comment

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
tpacuszkapl: It looks nice, but typescript complains. It has inferred type `'scalaVersion'|'builder'|...` and won't allow plain `string` :(.

Issue #133, opened 2022-08-15
View project contents without downloading the project


adamw:
As in start.spring.io's `explore`

https://twitter.com/Anh32920995/status/1559206242329001986

adamw:
Might require https://github.com/softwaremill/adopt-tapir/issues/126

PR #158, opened 2022-09-01
View project contents without downloading the project - backend

rafalambrozewicz:
These changes partly solve #133 as they are only on the backend side.

* Added `content` endpoint, that returns the structure of the generated project, so it could be displayed later on the frontend side.
* Introduced  ADT with `File` / `Directory` data types to enable representing the project as a tree of files.
* Added "project as a list of generated files" (`List[GeneratedFile]`) to "tree structure" (`Node`) conversion logic with its tests.
* Changed the way project files are formatted by scalafmt, now only the conf file gets written to disk (as it is required by the lib) and formatting of the content of the files is performed in memory.
* Extracted "formatting" and "file" logic to separate objects.


backend/src/main/scala/com/softwaremill/adopttapir/infrastructure/Json.scala
```diff
@@ -1,12 +1,23 @@
...
 /** Import the members of this object when doing JSON serialisation or deserialisation.
   */
 object Json extends AutoDerivation {
+
+  implicit lazy val circeConfiguration: CirceConfiguration =
+    CirceConfiguration.default.withDiscriminator("type")
+      .copy(transformConstructorNames = (s => s.toLowerCase)
+    )
+
+  implicit val tapirConfiguration: TapirConfiguration =
```
adamw, line 17: shouldn't this go to `http`? It doesn't have to do anything with json serialisation

backend/src/main/scala/com/softwaremill/adopttapir/infrastructure/Json.scala
```diff
@@ -1,12 +1,23 @@
...
-import io.circe.generic.AutoDerivation
 import io.circe.{Decoder, Encoder, Printer}
+import io.circe.generic.extras.{AutoDerivation, Configuration => CirceConfiguration}
+import sttp.tapir.generic.{Configuration => TapirConfiguration}
 
 /** Import the members of this object when doing JSON serialisation or deserialisation.
   */
 object Json extends AutoDerivation {
+
+  implicit lazy val circeConfiguration: CirceConfiguration =
```
adamw, line 12: no need for `lazy`

backend/src/main/scala/com/softwaremill/adopttapir/infrastructure/Json.scala
```diff
@@ -1,12 +1,23 @@
...
+    CirceConfiguration.default.withDiscriminator("type")
+      .copy(transformConstructorNames = (s => s.toLowerCase)
+    )
+
+  implicit val tapirConfiguration: TapirConfiguration =
+    TapirConfiguration.default.withDiscriminator("type")
+      .copy(toDiscriminatorValue = _.fullName.split('.').last.stripSuffix("$").toLowerCase)
```
adamw, line 19: this looks like the default `Configuration:51`?

backend/src/main/scala/com/softwaremill/adopttapir/infrastructure/Json.scala
```diff
@@ -1,12 +1,23 @@
...
   */
 object Json extends AutoDerivation {
+
+  implicit lazy val circeConfiguration: CirceConfiguration =
+    CirceConfiguration.default.withDiscriminator("type")
+      .copy(transformConstructorNames = (s => s.toLowerCase)
+    )
+
+  implicit val tapirConfiguration: TapirConfiguration =
+    TapirConfiguration.default.withDiscriminator("type")
```
adamw, line 18: would be good to extract hte shared discriminator name to a value

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
```diff
@@ -0,0 +1,51 @@
+package com.softwaremill.adopttapir.starter.content
+
+import com.softwaremill.adopttapir.logging.FLogging
+import com.softwaremill.adopttapir.starter.StarterDetails
+import com.softwaremill.adopttapir.template.ProjectGenerator
+
+class ContentService(projectGenerator: ProjectGenerator) extends FLogging {
```
adamw, line 7: logging isn't used?

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  def generateContentTree(starterDetails: StarterDetails): Node = {
+    val projectName = starterDetails.projectName
+    val projectAsFilesWithPaths = projectGenerator.generate(starterDetails)
+    val projectAsDirTrees = projectAsFilesWithPaths.map(gf => {
```
adamw, line 12: shouldn't we somehow delete/cleanup the generated files?

backend/src/test/scala/com/softwaremill/adopttapir/starter/content/ContentServiceTest.scala
```diff
@@ -0,0 +1,23 @@
+package com.softwaremill.adopttapir.starter.content
+
+import com.softwaremill.adopttapir.starter.{Setup, StarterDetails}
+import com.softwaremill.adopttapir.template.ProjectGenerator
+import com.softwaremill.adopttapir.test.BaseTest
+
+class ContentServiceTest extends BaseTest {
```
adamw, line 7: it would be good to have a test for one specific configuration checking something more than "it generated" - verifying some of the generated `Node` structure, content, etc. We aren't really testing anywhere that the generated `Node`s are correct

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
```diff
@@ -0,0 +1,51 @@
+package com.softwaremill.adopttapir.starter.content
```
adamw, line 1: it seems the code is not formatted - make sure that you have formatting enabled (should be automatic with intellij)

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
adamw: ah, these are just in-memory ones; but then we don't format these files

you'll probably need to extract appropriate code from `StarterService`

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
```diff
@@ -0,0 +1,51 @@
...
+class ContentService(projectGenerator: ProjectGenerator) extends FLogging {
+
+  def generateContentTree(starterDetails: StarterDetails): Node = {
+    val projectName = starterDetails.projectName
+    val projectAsFilesWithPaths = projectGenerator.generate(starterDetails)
+    val projectAsDirTrees = projectAsFilesWithPaths.map(gf => {
+      val pathsWithRootDir = (projectName + "/" + gf.relativePath).split('/').toList
+      val content = gf.content
+      createTree(pathsWithRootDir, content)
+    }).map(_.asInstanceOf[Directory])
```
adamw, line 16: why would you need that cast?

adamw: ah for the merge ... but what if we have top-level files? would be much better if this worked regardless of the file structure

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  private def merge(n1: Directory, n2: Directory): Directory = {
+    val files = n1.content.filter(_.isInstanceOf[File]) ++ n2.content.filter(_.isInstanceOf[File])
+    val n1Dirs = n1.content.filter(_.isInstanceOf[Directory]).map(_.asInstanceOf[Directory])
+    val n1DirsNames = n1Dirs.map(_.name)
+    val n2Dirs = n2.content.filter(_.isInstanceOf[Directory]).map(_.asInstanceOf[Directory])
+    val n2DirsNames = n2Dirs.map(_.name)
+
+    val n1UniqueDirs = n1Dirs.filter(n1d => !n2DirsNames.contains(n1d.name))
```
adamw, line 36: not sure what happens here but maybe it would be much simpler to create a `Set` of all directory names?

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
```diff
@@ -0,0 +1,51 @@
...
+    val n2UniqueDirs = n2Dirs.filter(n2d => !n1DirsNames.contains(n2d.name))
+    val uniqueDirs = n1UniqueDirs ++ n2UniqueDirs
+    val n1duplicatedDirs = n1Dirs.filter(n1d => n2DirsNames.contains(n1d.name))
+    val n2duplicatedDirs = n2Dirs.filter(n2d => n1DirsNames.contains(n2d.name))
+
+    val mergedDirs = n1duplicatedDirs.map(dd => {
+      merge(dd, n2duplicatedDirs.find(_.name == dd.name).get)
```
adamw, line 43: `.get` is another sign of potential problems (and exceptions, which we want to avoid)

I think once you have the set of directory names you can simply match on the result of `n1.children.find(_.name == n)` and the same with `n2`

backend/src/test/scala/com/softwaremill/adopttapir/starter/content/ContentServiceTest.scala
adamw: additionally we need to have a separate test for the (quite complicated) merging logic, and write some really good test cases there

probably the whole filepath+content -> node conversion can be extracted to an `object` or sth like that and tested independently of the starter service etc.

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
rafalambrozewicz: I am not sure if I've found the correct option - do you mean enabling something like "reformat code on save"?

backend/src/main/scala/com/softwaremill/adopttapir/infrastructure/Json.scala
rafalambrozewicz: yes, the only difference is to lower case conversion

adamw: then it's best to reuse existing by chaining functions (`Configuration.default.toEncodedName.andThen(...)`)

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
adamw: That's what I have:

image

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
rafalambrozewicz: Since top level files are not allowed, it would throw an exception - but sure, I'll try to make it more explicit and reduce number of casts.

backend/src/main/scala/com/softwaremill/adopttapir/infrastructure/Json.scala
rafalambrozewicz: it should, moved it there.

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
rafalambrozewicz: Removed `FLogging ` if not used.

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
rafalambrozewicz: Moved in-memory-files formatting to a different object, since it become not only `StarterService` responsibility.

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
rafalambrozewicz: Please look at the updated `merge` logic in `ContentMerger,` I hope it looks clearer.

backend/src/main/scala/com/softwaremill/adopttapir/http/Http.scala
```diff
@@ -13,10 +13,16 @@ import sttp.tapir.codec.enumeratum.TapirCodecEnumeratum
...
 import sttp.tapir.{Codec, Endpoint, EndpointOutput, PublicEndpoint, Schema, SchemaType, Tapir}
+import sttp.tapir.generic.{Configuration => TapirConfiguration}
 
 /** Helper class for defining HTTP endpoints. Import the members of this class when defining an HTTP API using tapir. */
 class Http() extends Tapir with TapirJsonCirce with TapirSchemas with TapirCodecEnumeratum with FLogging {
 
+  implicit val tapirConfiguration: TapirConfiguration =
+    TapirConfiguration.default
+      .withDiscriminator(Http.DiscriminatorName)
+      .copy(toDiscriminatorValue = TapirConfiguration.default.toEncodedName.compose(_.fullName.toLowerCase))
```
adamw, line 24: shouldn't this be `TapirConfiguration.default.toDiscriminatorValue` ?

backend/src/main/scala/com/softwaremill/adopttapir/infrastructure/Json.scala
```diff
@@ -1,12 +1,20 @@
...
 import com.softwaremill.tagging.@@
-import io.circe.generic.AutoDerivation
 import io.circe.{Decoder, Encoder, Printer}
+import io.circe.generic.extras.{AutoDerivation, Configuration => CirceConfiguration}
 
 /** Import the members of this object when doing JSON serialisation or deserialisation.
   */
 object Json extends AutoDerivation {
+
+  private val DiscriminatorName = "type"
```
adamw, line 11: the whole point in extracting this as a val was to define this in a single place and share among `Json`/`Http` :)

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentMerger.scala
```diff
@@ -0,0 +1,45 @@
+package com.softwaremill.adopttapir.starter.content
+
+object ContentMerger {
```
adamw, line 3: Maybe simply `DirectoryMerger` and an `apply` method? It's not quite clear what `create` does in an object called `Merger` :)

backend/src/main/scala/com/softwaremill/adopttapir/starter/files/FilesManager.scala
```diff
@@ -0,0 +1,50 @@
...
+import java.nio.file.Path
+import scala.reflect.io.Directory
+
+class FilesManager(config: StorageConfig) {
+
+  def createTempDir(): IO[File] = {
+    IO.blocking(newTemporaryDirectory(prefix = config.tempPrefix).toJava)
+  }
+
+  def createFiles(destinationDir: File, filesToCreate: List[GeneratedFile]): IO[List[Path]] = {
```
adamw, line 20: [minor] maybe we could consistently return either Path or File?

backend/src/main/scala/com/softwaremill/adopttapir/starter/formatting/ProjectFormatter.scala
```diff
@@ -0,0 +1,51 @@
...
+              tempDir <- tempDirectory
+              formatFile <- fm.createFile(tempDir, formatFile)
+              formattedFile <- IO(formatScalaFiles(formatFile, gfs))
+            } yield formattedFile
+
+            formattedFiles
+          }(release = tempDirectory => fm.deleteFilesAsStatedInConfig(tempDirectory))
+      }
+      case None => {
+        logger.error(s"Cannot find formatting file in generated project, scala files would NOT be formatted!")
```
adamw, line 32: [minor] **will** not would :)

backend/src/main/scala/com/softwaremill/adopttapir/starter/formatting/ProjectFormatter.scala
```diff
@@ -0,0 +1,51 @@
...
+  private val ScalaFileExtensions = Set(".scala", ".sbt")
+
+  private lazy val scalafmt = Scalafmt.create(getClass.getClassLoader)
+
+  def format(gfs: List[GeneratedFile]): IO[List[GeneratedFile]] = {
+    findGeneratedFormatFile(gfs) match {
+      case Some(formatFile) => {
+        IO.blocking(fm.createTempDir())
+          .bracket { tempDirectory =>
+            val formattedFiles = for {
```
adamw, line 22: [minor] no need for the assignment, you can just return the result of the `for` directly

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentMerger.scala
```diff
@@ -0,0 +1,45 @@
...
+    def innerCreateTree(pathNames: List[String], content: String): Node = {
+      pathNames match {
+        case Nil          => throw new IllegalStateException("pathNames list cannot be empty!")
+        case last :: Nil  => File(last, content)
+        case head :: tail => Directory(head, List(innerCreateTree(tail, content)))
+      }
+    }
+
+    val pathsWithRootDir = List(topDirName) ++ pathNames
+    innerCreateTree(pathsWithRootDir, content).asInstanceOf[Directory]
```
adamw, line 17: Instead of `.asInstanceOf`, maybe perform a `match` throwing an meaningful exception if a top-level file is found?

adamw:
The service structure and merger code look much better :)

backend/src/main/scala/com/softwaremill/adopttapir/starter/formatting/ProjectFormatter.scala
```diff
@@ -0,0 +1,51 @@
...
+    }
+  }
+
+  private def findGeneratedFormatFile(gfs: List[GeneratedFile]): Option[GeneratedFile] = {
+    gfs.find(_.relativePath.endsWith(CommonObjectTemplate.scalafmtConfigPath))
+  }
+
+  private def formatScalaFiles(scalaFormatPath: Path, gfs: List[GeneratedFile]): List[GeneratedFile] = {
+    gfs.map {
+      case gf if isScalaFilePath(gf.relativePath) =>
```
geminicaprograms, line 44: nit: naming - `gf` would be self explanatory if it was turned to either `generatedFile` or just simply `file`

backend/src/test/scala/com/softwaremill/adopttapir/starter/FileOperation.scala
```diff
@@ -2,13 +2,20 @@ package com.softwaremill.adopttapir.starter
...
 object FileOperation extends IOApp {
 
-  private val cfg = Config.read.starter
-  val service = new StarterService(cfg.copy(deleteTempFolder = false), new ProjectGenerator())
+  val service: StarterService = {
+    val cfg = Config.read.storageConfig.copy(deleteTempFolder = false)
+    val pg = new ProjectGenerator()
+    val fm = new FilesManager(cfg)
+    val pf = new ProjectFormatter(fm)
+    new StarterService(pg, fm, pf)
```
geminicaprograms, line 17: nit: naming - abbreviations are harder to follow as one needs to remember what they are standing for simple `projectGenerator`, `filesManager`, `filesFormatter` would do

backend/src/main/scala/com/softwaremill/adopttapir/starter/formatting/ProjectFormatter.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import cats.effect.IO
+import com.softwaremill.adopttapir.logging.FLogging
+import com.softwaremill.adopttapir.starter.files.FilesManager
+import com.softwaremill.adopttapir.template.{CommonObjectTemplate, GeneratedFile}
+import org.scalafmt.interfaces.Scalafmt
+
+import java.nio.file.Path
+
+class ProjectFormatter(fm: FilesManager) extends FLogging {
```
geminicaprograms, line 11: nit: naming AFAIU it is used to format files in the project therefore IMHO `FilesFormatter` would better reflect its purpose

backend/src/test/scala/com/softwaremill/adopttapir/starter/content/ContentServiceTest.scala
```diff
@@ -0,0 +1,32 @@
...
+      new ContentService(pg, pf)
+    }
+  }
+
+  import ContentServiceTest._
+
+  it should "generate project tree for every valid configuration" in {
+    allStarterDetails().foreach(sd => {
+      service.generateContentTree(sd)
+      // if it would throw, the test would fail
```
geminicaprograms, line 27: how about:
`The content tree generation failure results in the exception being thrown and as a result a test failure`

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentMerger.scala
rafalambrozewicz: This is a very nice suggestion. Thanks! 👍

backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterService.scala
```diff
@@ -1,49 +1,35 @@
...
-    destination.toJava
+  def generateZipFile(starterDetails: StarterDetails): IO[File] = {
+    val rawGeneratedFiles = pg.generate(starterDetails)
+    val formattedGeneratedFiles = pf.format(rawGeneratedFiles)
+
+    val zipFile = IO
+      .blocking(fm.createTempDir())
+      .bracket { tempDirectory =>
+        for {
+          tempDir <- tempDirectory
```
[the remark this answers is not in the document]
rafalambrozewicz: I think that, it is. In this case `tempDirectory` is a `IO[File]` and `tempDir` is a `File`, that would be used in later calls.

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentMerger.scala
```diff
@@ -0,0 +1,45 @@
...
+  def createTree(topDirName: String, pathNames: List[String], content: String): Directory = {
+    assert(pathNames.nonEmpty, "`pathNames` list cannot be empty!")
+
+    def innerCreateTree(pathNames: List[String], content: String): Node = {
+      pathNames match {
+        case Nil          => throw new IllegalStateException("pathNames list cannot be empty!")
+        case last :: Nil  => File(last, content)
+        case head :: tail => Directory(head, List(innerCreateTree(tail, content)))
+      }
+    }
```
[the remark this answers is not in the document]
rafalambrozewicz: You are right, `innerCreateTree` could be defined outside `createTree` as a private method. There are no obvious gains here, I've used to use such inner methods when I need to do something before I could invoke a recursive call. Are there any drawbacks to using such an approach?

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
```diff
@@ -0,0 +1,24 @@
...
+
+class ContentService(pg: ProjectGenerator, pf: ProjectFormatter) {
+
+  def generateContentTree(starterDetails: StarterDetails): IO[Node] = {
+    val projectName = starterDetails.projectName
+    val rawGeneratedFiles = pg.generate(starterDetails)
+
+    for {
+      formattedGeneratedFiles <- pf.format(rawGeneratedFiles)
+      projectAsDirTrees <- IO(formattedGeneratedFiles.map(fgf => {
```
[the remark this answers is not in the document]
rafalambrozewicz: I would rather leave it as it is - the scope where `fgf` is used is relatively small, at worst one glance 2 lines above explains everything (i.e. `fgf` is an element of `formattedGeneratedFiles` collection).

backend/src/test/scala/com/softwaremill/adopttapir/starter/FileOperation.scala
rafalambrozewicz: I would rather leave it as it is - please notice, that all of these are used in stricly limited scope in order to create `StarterService`.

backend/src/main/scala/com/softwaremill/adopttapir/starter/formatting/ProjectFormatter.scala
rafalambrozewicz: It's a bit more complicated. We format something called `GeneratedFile` - entities that consist of `path` and `content` strings. Renaming `ProjectFormatter` to `FilesFormatter` implies that `File`s are involved and that is not true. Perhaps `GeneratedFilesFormatter` would be a better name, but it is quite long name...

backend/src/main/scala/com/softwaremill/adopttapir/infrastructure/Json.scala
rafalambrozewicz: Moved `DiscriminatorName ` to `Constants`

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentMerger.scala
```diff
@@ -0,0 +1,45 @@
...
+
+    val pathsWithRootDir = List(topDirName) ++ pathNames
+    innerCreateTree(pathsWithRootDir, content).asInstanceOf[Directory]
+  }
+
+  def merge(d1: Directory, d2: Directory): Directory = {
+    assert(d1.name == d2.name, "names of directories to merge cannot be different!")
+
+    val (d1DirsAll, d2DirsAll) = (d1.content.collect { case d: Directory => d }.toSet, d2.content.collect { case d: Directory => d }.toSet)
```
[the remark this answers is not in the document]
rafalambrozewicz: Added `childDirectories()` and `childFiles()`. Thanks!

backend/src/main/scala/com/softwaremill/adopttapir/starter/files/FilesManager.scala
```diff
@@ -0,0 +1,50 @@
...
+    destination.toJava
+  }
+
+  def deleteFilesAsStatedInConfig(destinationDir: IO[File]): IO[Unit] = IO.blocking {
+    if (config.deleteTempFolder) {
+      for {
+        dir <- destinationDir
+        _ <- IO(new Directory(dir).deleteRecursively())
+      } yield dir
+      ()
```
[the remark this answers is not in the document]
rafalambrozewicz: I've thought, that I need to explicitly return unit `()` here, but I do not, so it would be removed.

backend/src/main/scala/com/softwaremill/adopttapir/starter/files/FilesManager.scala
```diff
@@ -0,0 +1,50 @@
...
+    } else {
+      ()
+    }
```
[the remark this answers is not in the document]
rafalambrozewicz: It is not needed. Thanks!

backend/src/main/scala/com/softwaremill/adopttapir/starter/formatting/ProjectFormatter.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  private lazy val scalafmt = Scalafmt.create(getClass.getClassLoader)
+
+  def format(gfs: List[GeneratedFile]): IO[List[GeneratedFile]] = {
+    findGeneratedFormatFile(gfs) match {
+      case Some(formatFile) => {
+        IO.blocking(fm.createTempDir())
+          .bracket { tempDirectory =>
+            val formattedFiles = for {
+              tempDir <- tempDirectory
```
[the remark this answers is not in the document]
rafalambrozewicz: It think it is since `tempDirectory` is `IO[File]` and `tempDir` is `File`

backend/src/main/scala/com/softwaremill/adopttapir/starter/formatting/ProjectFormatter.scala
```diff
@@ -0,0 +1,51 @@
...
+          }(release = tempDirectory => fm.deleteFilesAsStatedInConfig(tempDirectory))
+      }
+      case None => {
+        logger.error(s"Cannot find formatting file in generated project, scala files would NOT be formatted!")
+        IO(gfs)
+      }
+    }
+  }
+
+  private def findGeneratedFormatFile(gfs: List[GeneratedFile]): Option[GeneratedFile] = {
```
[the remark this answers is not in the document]
rafalambrozewicz: Very limited scope. I would leave it as it is.

backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterService.scala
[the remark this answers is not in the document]
geminicaprograms: right, haven't noticed that `tempDirectory` is of `IO[...]` type.

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentMerger.scala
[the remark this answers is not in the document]
geminicaprograms: It can be reviewed/examined in isolation (which is good cuz it is a method) yet it increases the complexity of `createTree` (caller). One can just put it before the caller so that it is visually _close_ to it (it answers the need of having definition just right before the code) and doesn't require much scrolling if needed...

backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterService.scala
```diff
@@ -1,49 +1,35 @@
...
-  }
-
-  private def storeFiles(destinationFolder: File, filesToCreate: List[GeneratedFile]): IO[Unit] = IO.blocking {
-    filesToCreate.foreach { generatedFile =>
-      val file = BFile(destinationFolder.getPath, generatedFile.relativePath)
-      file.parent.createDirectories()
-      file.overwrite(generatedFile.content)
-    }
-  }
+class StarterService(pg: ProjectGenerator, fm: FilesManager, pf: ProjectFormatter) extends FLogging {
```
geminicaprograms, line 13: as we discussed - it is easier to follow class members when they have not abbreviated names; e.g. it used to be `projectGenerator` for `pg` etc...

backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterService.scala
```diff
@@ -1,49 +1,35 @@
...
 
-  private def zipDirectory(directoryFile: File): IO[File] = IO.blocking {
-    val destination = BFile.newTemporaryFile(prefix = directoryFile.getName + "_", suffix = ".zip")
-    ZipArchiver().create(destination.path, directoryFile.toPath)
-    destination.toJava
+  def generateZipFile(starterDetails: StarterDetails): IO[File] = {
+    val rawGeneratedFiles = pg.generate(starterDetails)
```
geminicaprograms, line 16: there is log statement missing
```
logger.info(s"received request: $starterDetails") *>
```
if we keep it then files generation should be wrapped with IO effect so that all happens on IO thread...

backend/src/main/scala/com/softwaremill/adopttapir/starter/files/FilesManager.scala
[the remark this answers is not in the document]
geminicaprograms: So can it be removed or not? ;)

backend/src/main/scala/com/softwaremill/adopttapir/starter/files/FilesManager.scala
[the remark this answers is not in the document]
geminicaprograms: So can it be removed or not? ;)

backend/src/main/scala/com/softwaremill/adopttapir/starter/formatting/ProjectFormatter.scala
geminicaprograms: +1 to `GeneratedFilesFormatter`

backend/src/test/scala/com/softwaremill/adopttapir/starter/FileOperation.scala
geminicaprograms: Ok, so maybe we don't need all of them?
How about
```
val fm = new FilesManager(cfg)
new StarterService(new ProjectGenerator(), fm, new ProjectFormatter(fm))
```
?

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentMerger.scala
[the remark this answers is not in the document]
rafalambrozewicz: Moved `innerCreateTree` outside `createTree` as private `createTreeRec`

backend/src/main/scala/com/softwaremill/adopttapir/starter/files/FilesManager.scala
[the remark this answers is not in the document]
rafalambrozewicz: Not, it is removed ;-)

backend/src/main/scala/com/softwaremill/adopttapir/starter/files/FilesManager.scala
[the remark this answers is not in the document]
rafalambrozewicz: Now, it is removed ;-)

backend/src/main/scala/com/softwaremill/adopttapir/starter/formatting/ProjectFormatter.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  def format(gfs: List[GeneratedFile]): IO[List[GeneratedFile]] = {
+    findGeneratedFormatFile(gfs) match {
+      case Some(formatFile) => {
+        IO.blocking(fm.createTempDir())
+          .bracket { tempDirectory =>
+            val formattedFiles = for {
+              tempDir <- tempDirectory
+              formatFile <- fm.createFile(tempDir, formatFile)
+              formattedFile <- IO(formatScalaFiles(formatFile, gfs))
```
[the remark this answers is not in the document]
rafalambrozewicz: Good catch. Updated name.

backend/src/main/scala/com/softwaremill/adopttapir/starter/formatting/ProjectFormatter.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  private def formatScalaFiles(scalaFormatPath: Path, gfs: List[GeneratedFile]): List[GeneratedFile] = {
+    gfs.map {
+      case gf if isScalaFilePath(gf.relativePath) =>
+        gf.copy(content = scalafmt.format(scalaFormatPath, Path.of(gf.relativePath), gf.content))
+      case gf => gf
+    }
+  }
+
+  private def isScalaFilePath(p: String): Boolean = ScalaFileExtensions.exists(e => p.endsWith(e))
```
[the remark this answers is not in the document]
rafalambrozewicz: Updated.

backend/src/test/scala/com/softwaremill/adopttapir/starter/FileOperation.scala
rafalambrozewicz: Updated.

backend/src/main/scala/com/softwaremill/adopttapir/starter/formatting/ProjectFormatter.scala
rafalambrozewicz: Renamed then.

backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterService.scala
rafalambrozewicz: Updated.

backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterService.scala
rafalambrozewicz: Updated. Log statement is now present.

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/DirectoryMerger.scala
```diff
@@ -0,0 +1,48 @@
+package com.softwaremill.adopttapir.starter.content
+
+object DirectoryMerger {
+
+  def createTree(topDirName: String, pathNames: List[String], content: String): Directory = {
```
adamw, line 5: I would insist that `createTree` does not belong in a `DirectoryMerger` object :) Instead, it looks like a separate `DirectoryCreator.apply` object

adamw (approved):

PR #159, opened 2022-09-02
Tailored scalafmt reporter.

micsza:


backend/src/main/scala/com/softwaremill/adopttapir/starter/FormatScalaFiles.scala
```diff
@@ -4,15 +4,25 @@ import better.files.FileExtensions
...
   private val ScalaFileExtensions = Set(".scala", ".sbt")
 
-  private lazy val scalafmt = Scalafmt.create(getClass.getClassLoader)
+  private lazy val scalafmt =
+    Scalafmt
+      .create(getClass.getClassLoader)
+      .withReporter(
+        new ConsoleScalafmtReporter(System.err) {
+          override def parsedConfig(config: Path, scalafmtVersion: String): Unit =
+            System.out.println(s"parsed config (v$scalafmtVersion): $config")
```
adamw, line 23: maybe we can do normal logging, `logger.info(...)`?

adamw:
@rafalambrozewicz I think this might conflict with your branch? Maybe it would be better to create the PR against it.

rafalambrozewicz:
> @rafalambrozewicz I think this might conflict with your branch? Maybe it would be better to create the PR against it.

Indeed, there would be a conflict, so it makes sense to merge this PR into `[133-view-project-contents-without-downloading-the-project](https://github.com/softwaremill/adopt-tapir/tree/133-view-project-contents-without-downloading-the-project)`

PR #162, opened 2022-09-05
fix: typos in the Scala CLI README.md template

geminicaprograms:
Apart from the obvious typos there refer to the product as `Scala CLI`
(the same way product is referred in its homepage).


adamw:
Thx :)

PR #165, opened 2022-09-09
Feat/share configuration

### LINKED ISSUE #75: [Feature] Introduce configuration sharing
Add `Share configuration` button (as highlighted below)
image

that generates link with all configuration passed as URL parameters. Configuration from screenshot would result in the link below
```
https://adopt-tapir.softwaremill.com?addDocumentation=true&addMetrics=true&effect=IOEffect&groupId=org.example&implementationHttp4s&json=Jsoniter&projectName=share
```

Note that clicking on `share` would add the URL to the clipboard.

geminicaprograms:
Add new `share` button that is displayed next to the `Generate tapir project' header. Note that upon clicking this button link (with the necessary query parameters) is copied to the clipboard (and confirmation is displayed).
When one uses the link it is validated and either applied (with a confirmation) or ignored with a warning that validation failed.

Here is the current look & feel with _copy to clipboard_ confirmation captured (note that when one hovers over the `share` button a tooltip is shown - also in this picture):
image

And here is the configuration successfully applied:
image

And here is the failure operation:
image
In such case default settings are applied.

Note that query parameters are stripped away from the URL after they get applied (regardless if it was success or failure).

Closes #75


ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
```diff
@@ -119,13 +165,53 @@ export const ConfigurationForm: React.FC<ConfigurationFormProps> = ({ isEmbedded
...
+          </Typography>
+          <Tooltip title="Share configuration" arrow>
+            <IconButton color="secondary" aria-label="share configuration" onClick={handleShareConfiguration}>
+              <ShareTwoToneIcon />
+            </IconButton>
+          </Tooltip>
+        </Stack>
       )}
+
+      <Snackbar
```
tpacuszkapl, line 204: There already is one snackbar on this page - in the `ApiCallAddons` component. Maybe it can be reused? Or at least unified, because this one differs a little (f.e. different horizontal position).

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
```diff
@@ -119,13 +165,53 @@ export const ConfigurationForm: React.FC<ConfigurationFormProps> = ({ isEmbedded
...
+    setSnackbar({ open: true, severity: 'info', message: 'Link to configuration copied to clipboard.' });
+  };
+
   return (
     <Box>
       {!isEmbedded && (
-        <Typography variant="h3" component="h3" fontWeight={300} gutterBottom>
-          Generate tapir project
-        </Typography>
+        <Stack direction="row" spacing={0.25}>
```
tpacuszkapl, line 192: Have you considered extracting this tag and its contents and functionality to separate component?

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
```diff
@@ -119,13 +165,53 @@ export const ConfigurationForm: React.FC<ConfigurationFormProps> = ({ isEmbedded
...
+          <Typography variant="h3" component="h3" fontWeight={300} gutterBottom>
+            Generate tapir project
+          </Typography>
+          <Tooltip title="Share configuration" arrow>
+            <IconButton color="secondary" aria-label="share configuration" onClick={handleShareConfiguration}>
```
tpacuszkapl, line 197: Nit: When I have hovered over this icon I saw this:
Screenshot from 2022-09-09 17-02-16
This darker background in ellipsis looks a little strange to me. Maybe circle would be better? Or changing not background of the icon, but icon color (f.e. from grey to orange?).

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
```diff
@@ -23,19 +24,64 @@ import {
...
+        .catch(_ => {
+          setSnackbar({
+            open: true,
+            severity: 'warning',
+            message: 'Validation of linked configuration failed therefore it was not applied.',
+          });
+        });
+      setSearchParams({});
+    }
+  });
```
tpacuszkapl, line 83: Hooks usually have dependency arrays, to run effects only when elements of those dependency arrays change (ref: https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects). In the past I've encountered subtle bugs when I left it off. But I'm not sure how important it is here, as you have if statements that prevents doing too much work here. What do you think @kkolasinski, should dependency array be filled here?

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
```diff
@@ -119,13 +165,53 @@ export const ConfigurationForm: React.FC<ConfigurationFormProps> = ({ isEmbedded
...
 
+  const [snackbar, setSnackbar] = useState<SnackbarConfig>({
+    open: false,
+    severity: 'info',
+  });
+  const handleSnackClose = (event?: SyntheticEvent | Event, reason?: string) => {
+    if (reason === 'clickaway') {
+      return;
+    }
+    setSnackbar({ open: false });
```
tpacuszkapl, line 176: Nit: this sets severity to `undefined` and `undefined` has color green. So when it fades, it flickers for a miliseconds from blue to green. Doing `setSnackbar({...snackbar, open: false})` should prevent it.

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
geminicaprograms: there was a dependency array filled with `setSearchParams` but `yarn lint` complained with the following
```
83:6  warning  React Hook useEffect has missing dependencies: 'form' and 'searchParams'. Either include them or remove the dependency array  react-hooks/exhaustive-deps
```
so I have removed it altogether... having said that I'm opened for suggestions ;)

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
geminicaprograms: Good catch! I haven't noticed that. Fixed.

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
geminicaprograms: I was rather thinking about componentizing `Tooltip` but I would like to do that as a follow up rather than here

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
geminicaprograms: I have no idea how to fix it but it bothered me too. @kkolasinski any suggestions would be welcomed. Another thing to _fix_ is that it looks a bit misaligned IMHO the share button should be a few pixels up (again have no idea on how to do it by the book ;))

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
geminicaprograms: Right. For now I have aligned positions. I will think about aligning them as here it contains also `x` so that one can clean it up nicely. IMHO it is a subject of the follow-up

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
kkolasinski: AFAIK empty deps array causes the `useEffect` hook to be ran once.

From what I understand, the idea is to set the form values by getting them from the URL.

If we're not planning on updating the URL in real time, we should be fine with omitting the dependency array - the URL params are read once during the initial render, so we don't need to track any changes.

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
kkolasinski: This can easily be fixed by setting the `alignItems` prop on the `Stack` component.

`<Stack direction="row" alignItems="baseline" spacing={0.25}>`

The value `baseline` aligns the button to the line height of the heading, another option would be `center`, but that centers vertically based on the container height, which still looks like the button is misaligned.

geminicaprograms: @kkolasinski that was the power comment as it fixed both `ellipsis shape` and `alignement` issues :)

geminicaprograms: thanks!

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
geminicaprograms: That is the intention: perform the operation in question once when the form is displayed for the first time.

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
geminicaprograms: @tpacuszkapl we have discussed with @kkolasinski that componentization works nice with the other parts of that form (drop downs, or radio groups, etc...) as they are enclose independent parts and they are re-used. However neither header nor `share-button` are subject of the same as they are not going to be re-used in the near future (there are no plans to have more forms here) and the `share-button` heavily depends on the form content so it kind of makes sense to keep it there... IOW unless you have another angle to view on it (that we are missing) I would refrain from componentizing it...

kkolasinski (approved):

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
tpacuszkapl: Fine with me

tpacuszkapl:
One more thing - I've clicked the button, got the link, pasted it into the browser, and after loading the page the share button is inactive. But I haven't touched the form, so it is still valid. 
It gets active after i click any input at click away from it. Could you check that?

geminicaprograms:
> One more thing - I've clicked the button, got the link, pasted it into the browser, and after loading the page the share button is inactive. But I haven't touched the form, so it is still valid. It gets active after i click any input at click away from it. Could you check that?

Done, thanks for catching it up!

tpacuszkapl (approved):

PR #178, opened 2022-09-14
Update to scala 3.2.0

micsza:


rafalambrozewicz (commented):
Great job, +1 for marking several classes, that we would not extend as `final`, also as emus got introduced in Scala 3 these parts look much cleaner now. Now, if I may grumble - let's use scalafm on all scala files - there as places with more than one space between words or that have multiple new lines at the end of the file, scalafm hopefully would take care of that. Additionally, I am not sure about using `:` instead of `{}`, especially in the case of bigger files, where I find bracket notation better as it is easier for me to distinguish given blocs of code... still - it is mostly personal preference. Kudos!

backend/src/main/scala/com/softwaremill/adopttapir/util/Util.scala
```diff
@@ -0,0 +1,8 @@
+package com.softwaremill.adopttapir.util
```
adamw, line 1: since this is not a class, I think the file name should be lower-cased

build.sbt
```diff
@@ -194,16 +191,29 @@ lazy val rootProject = (project in file("."))
...
   .aggregate(backend, ui, templateDependencies)
 
-lazy val ItTest = config("ItTest") extend (Test)
+lazy val ItTest = config("ItTest") extend Test
 
 def itFilter(name: String): Boolean = name endsWith "ITTest"
 def unitFilter(name: String): Boolean = (name endsWith "Test") && !itFilter(name)
 
+val backendExtraScalacOptions = Seq(
+  "-indent",
```
adamw, line 200: do we still need this?

build.sbt
```diff
@@ -9,20 +9,22 @@ import sbtbuildinfo.{BuildInfoKey, BuildInfoOption}
 import scala.sys.process.Process
 import scala.util.Try
 
-val tapirVersion = "1.1.2"
+val scala2Version = "2.13.10"
+val scala32Version = "3.2.0"
```
adamw, line 13: simply `scala3Version`?

backend/src/main/scala/com/softwaremill/adopttapir/config/Config.scala
```diff
@@ -4,16 +4,15 @@ import com.softwaremill.adopttapir.http.HttpConfig
 import com.softwaremill.adopttapir.starter.files.StorageConfig
 import com.softwaremill.adopttapir.version.BuildInfo
 import com.typesafe.scalalogging.StrictLogging
-import pureconfig.ConfigSource
-import pureconfig.generic.auto._
```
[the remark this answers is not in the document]
adamw: I suspect this https://pureconfig.github.io/docs/scala-3-derivation.html doesn't work? :)

backend/src/main/scala/com/softwaremill/adopttapir/config/ConfigReader.scala
```diff
@@ -0,0 +1,45 @@
...
+import com.softwaremill.adopttapir.starter.api.StarterApi
+import com.softwaremill.adopttapir.starter.content.ContentService
+import com.softwaremill.adopttapir.starter.files.FilesManager
+import com.softwaremill.adopttapir.starter.formatting.GeneratedFilesFormatter
+import com.softwaremill.adopttapir.template.ProjectGenerator
+import sttp.tapir.server.metrics.prometheus.PrometheusMetrics
+
+type ConfigReader[A] = Reader[Config, A]
+
+object ConfigReader:
```
adamw, line 19: I expected this to read the config file, but in fact this does something else - it performs the wiring of dependencies? I think this might go to `Dependencies` as private members? `Reader` is more of a technical detail, than the main responsibility of the class :)

backend/src/main/scala/com/softwaremill/adopttapir/Dependencies.scala
```diff
@@ -2,39 +2,23 @@ package com.softwaremill.adopttapir
...
+import com.softwaremill.adopttapir.config.{Config, ConfigReader}
 import com.softwaremill.adopttapir.http.{Http, HttpApi, HttpConfig}
 import com.softwaremill.adopttapir.metrics.VersionApi
 import com.softwaremill.adopttapir.starter.api.StarterApi
-import com.softwaremill.macwire.autocats.autowire
 import sttp.tapir.server.metrics.prometheus.PrometheusMetrics
 
 case class Dependencies(api: HttpApi)
 
 object Dependencies {
```
adamw, line 14: shouldn't we use `:` here as well?

backend/src/main/scala/com/softwaremill/adopttapir/starter/formatting/GeneratedFilesFormatter.scala
```diff
@@ -2,14 +2,18 @@ package com.softwaremill.adopttapir.starter.formatting
...
 import org.scalafmt.interfaces.{Scalafmt, ScalafmtReporter}
 
 import java.io.{OutputStreamWriter, PrintWriter}
 import java.nio.file.Path
 
-class GeneratedFilesFormatter(fm: FilesManager) extends StrictLogging {
+final case class GeneratedFilesFormatter(storageConfig: StorageConfig) extends StrictLogging:
+
+  private val filesManager: FilesManager = FilesManager(storageConfig)
+  export filesManager.*
```
adamw, line 16: why would a class whose responsibility is formatting, include methods for managing files? shouldn't these remain separate?

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentModel.scala
```diff
@@ -1,11 +1,29 @@
...
+import sttp.tapir.Schema
+import io.circe.*
+
+sealed trait Node: // derives Codec.AsObject:
   def name: String
-}
 
-case class File(name: String, content: String) extends Node
-case class Directory(name: String, content: List[Node]) extends Node {
+object Node:
```
adamw, line 15: I see you've moved from defining all encoders/decoders externally (as part of `Json`), to defining them as part of the ADT. This works here, since both are in the same sub-project, but I think if we do so, we should do it consistently. For example, `Schema` derivation should be included here as well (instead of being part of `Tapir`). Plus sometimes, the encoders/decoders are defined as implicit values, and sometimes using `derives`.

backend/src/main/scala/com/softwaremill/adopttapir/config/Config.scala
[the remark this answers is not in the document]
micsza: Partially it does work - refactored.

backend/src/main/scala/com/softwaremill/adopttapir/util/Util.scala
micsza: Renamed.

build.sbt
micsza: Nope, removed.

build.sbt
micsza: Sure.

backend/src/main/scala/com/softwaremill/adopttapir/config/ConfigReader.scala
micsza: Refactored.

backend/src/main/scala/com/softwaremill/adopttapir/Dependencies.scala
micsza: Sure.

backend/src/main/scala/com/softwaremill/adopttapir/starter/formatting/GeneratedFilesFormatter.scala
micsza: That was a failed try to simplify things but yes defo formatter should have been separated from manager and then eventually some facade built on top of both. Anyway, I reversed back to previous solution.

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentModel.scala
micsza: It's mostly due to some limitations of `circe 0.14` for `Scala 3`. I have included `Schema` derivation in case classes but I think I need to keep the `implicit` instance of `Node` due to https://tapir.softwaremill.com/en/latest/endpoint/schemas.html#derivation-for-recursive-types-in-scala3 : when deriving `Schema` for `Node` (i.e. no `implicit` instance of `Schema[Node]`), it compiles but then seems to fall in endless loop at runtime.

backend/src/main/scala/com/softwaremill/adopttapir/starter/formatting/GeneratedFilesFormatter.scala
adamw: thanks :)

backend/src/main/scala/com/softwaremill/adopttapir/config/Config.scala
```diff
@@ -29,7 +28,8 @@ object Config extends StrictLogging {
     }
 
     logger.info(info)
-  }
+
+  given ConfigReader[Config] =
```
adamw, line 32: no `derives`? :)

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentModel.scala
micsza: Comment added.

backend/src/main/scala/com/softwaremill/adopttapir/config/Config.scala
micsza: Fixed.

micsza:
@rafalambrozewicz 
Many thanks :) Formatting applied. As for `{` vs `:` - well a matter of taste ;) As a compromise, I have closed the scopes explicitly with `end` where I found it might increase readibility.

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentModel.scala
```diff
@@ -1,11 +1,34 @@
...
+  implicit def schema: Schema[Node] = Schema.derived[Node]
+
+  given Encoder[Node] = Encoder.instance {
+    case file @ File(_, _)     => file.asJson.mapObject(_.add("type", "file".asJson))
+    case dir @ Directory(_, _) => dir.asJson.mapObject(_.add("type", "directory".asJson))
+  }
+
+  given Decoder[Node] = Decoder[File].or(Decoder[Directory].widen)
+
+case class File(name: String, content: String) extends Node derives Codec.AsObject, Schema
```
adamw, line 30: here we add `derives ... Schema`, but e.g. `Error_OUT` doesn't have this. Shouldn't it? Maybe we have somewhere `import sttp.tapir.generic.auto._` which would auto-derive schemas when it shouldn't

backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequest.scala
```diff
@@ -13,113 +15,32 @@ case class StarterRequest(
...
-  }
-
-  override def values: IndexedSeq[BuilderRequest] = findValues
-}
+) derives Decoder
+
+object StarterRequest:
+  inline given Encoder[StarterRequest] = deriveEncoder
+
+enum EffectRequest(val toModel: ServerEffect) derives JsonTaggedAdt.PureEncoder, JsonTaggedAdt.PureDecoder:
```
adamw, line 23: maybe also add a comment why we need `JsonTaggedAdt` over the usual `derives Decoder`

backend/src/main/scala/com/softwaremill/adopttapir/metrics/VersionApi.scala
```diff
@@ -2,24 +2,27 @@ package com.softwaremill.adopttapir.metrics
...
     }
-}
 
-object VersionApi {
-  case class Version_OUT(buildSha: String)
-}
+object VersionApi:
+  final case class Version_OUT(buildSha: String) derives Decoder
+
+  object Version_OUT:
```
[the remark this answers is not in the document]
micsza: Sure it can ;). Changed.

backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequest.scala
micsza: Added.

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentModel.scala
micsza: We are not using `import sttp.tapir.generic.auto._`. There is though that `TapirSchemas` trait that brings up the Schema givens inside `Http` class.

adamw: Well `trait TapirSchemas extends SchemaDerivation` which is auto-derivation. So I think we should remove that `... extends SchemaDerivation` and see if things compile.

backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequest.scala
```diff
@@ -13,113 +15,34 @@ case class StarterRequest(
     json: JsonImplementationRequest,
     scalaVersion: ScalaVersionRequest,
     builder: BuilderRequest
-)
+) derives Decoder
 
-object StarterRequest {}
+object StarterRequest:
+  inline given Encoder[StarterRequest] = deriveEncoder
```
adamw, line 21: shouldn't this be moved to `derives`?

backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequest.scala
```diff
@@ -13,113 +15,34 @@ case class StarterRequest(
...
-  override def values: IndexedSeq[JsonImplementationRequest] = findValues
-}
-
-sealed trait ScalaVersionRequest extends EnumEntry {
-  def toModel: ScalaVersion
-}
-
-/** Changes in Scala versions have to be reflected in .github/workflows/adopt-tapir-ci.yml file so that running jobs in parallel is still
-  * possible
+/** We use JsonTaggedAdt to serialize starter request enums in a pure, tags-constant only way.
```
adamw, line 23: ok, but why `JsonTaggedAdt` and not `Encoder` simply?

adamw: (that's what I'd put in the comment)

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentModel.scala
micsza: I have removed the whole `TapirSchema` trait and added `derives ... Schema` in few more places.

backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequest.scala
micsza: Yes, it should. Changed.

backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequest.scala
micsza: Hopefully, it's somehow improved now.

backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequest.scala
```diff
@@ -13,113 +16,36 @@ case class StarterRequest(
...
-  def toModel: JsonImplementation
-}
-
-/** Changes in JSON implementation have to be reflected in .github/workflows/adopt-tapir-ci.yml file so that running jobs in parallel is
-  * still possible
+) derives Decoder,
+      Encoder.AsObject,
+      Schema
+
+/** Here I deviate from default circe marshallers to a specialized JsonTaggedAdt to serialize enums in a tags-constant only way. The reason
```
adamw, line 23: Thanks! :) all clear now

adamw:
Done! :) Let's see how this works now :)

PR #179, opened 2022-09-14
feat: preview files ux improvements

### LINKED ISSUE #170: Preview: UX issues
* the generated code & readme need to be side-scrolled: add line breaks or decrease column width in the formatter when previewing so that it doesn't happen
* copying code from preview also copies line numbers; only code should be copied
* maybe all directories should be expanded by default, and the most interesting file - Main.scala - selected?

geminicaprograms:
The following improvements were added:
* _preview_ layout was improved to offer more space for the editor (files tree was narrowed, code editor was expanded, margins got thinner)
* all `src` folders are opened
* `Main.scala` file content is selected in the editor
* folder representation was modified to slightly visually distinguish between folder closed/opened states

Here is the look & feel of `preview`:
image

Closes #170

### DIFF
--- a/ui/src/components/FileTreeView/FileTreeView.types.ts
+++ b/ui/src/components/FileTreeView/FileTreeView.types.ts
@@ -23,3 +23,20 @@ export type TreeState = {
   openedFile: NodeAbsoluteLocation;
   openFile: (loc: NodeAbsoluteLocation) => void;
 };
+
+export function getAllDirectories(prefix: string[], files: FileTreeNode[]): string[][] {
+  const dirs: string[][] = [];
+  files
+    .filter(isDirectory)
+    .map(file => file as DirNode)
+    .forEach(dir => {
+      const newPrefix = [...prefix, dir.name];
+      dirs.push(newPrefix);
+      dirs.push(...getAllDirectories(newPrefix, dir.content));
+    });
+  return dirs;
+}
+
+const isDirectory = (node: FileTreeNode): boolean => {
+  return node.type === 'directory';
+};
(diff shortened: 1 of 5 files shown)


ui/src/components/FileTreeView/FileTreeView.types.ts
```diff
@@ -23,3 +23,69 @@ export type TreeState = {
...
+    .map(file => file as DirNode)
+    .forEach(dir => {
+      const newPrefix = [...prefix, dir.name];
+      dirs.push(newPrefix);
+      dirs.push(...getAllDirectories(newPrefix, dir.content));
+    });
+  return dirs;
+}
+
+export function getMainScala(files: FileTree): string[] {
```
tpacuszkapl, line 40: In the `FileTreeView` we have access to the `groupId`, maybe, instead of the search, we could split it and concatenate absolute path from it?

ui/src/pages/PreviewStarterPage/PreviewStarterPage.component.tsx
```diff
@@ -19,6 +26,25 @@ export function PreviewStarterPage() {
   const treeState = useTreeState('README.md');
   const { call, clearError, isLoading, errorMessage } = useApiCall();
 
+  // function references were introduced so that 'useCallback' can depend on them
+  const openFileRef = treeState.openFile;
+  const toggleDirRef = treeState.toggleDir;
+
+  const openMainScala = useCallback(() => {
```
tpacuszkapl, line 33: Is there a reason for using `useCallback`?
I would consider using `useEffect` here instead of creating a separate callback and calling it in another `useEffect`.

tpacuszkapl (commented):
Two minor suggestions

ui/src/pages/PreviewStarterPage/PreviewStarterPage.component.tsx
geminicaprograms: That was some historical reason which is no longer valid. Good catch I have just removed it!

ui/src/components/FileTreeView/FileTreeView.types.ts
geminicaprograms: Good idea. It simplifies the calculations and `gatMainScala` (and methods it depends on) can be removed.

tpacuszkapl (approved):

Issue #130, opened 2022-08-12
Tidy up the main UI form


adamw:
I think the form might use some grouping and tidying up as it became a bit chaotic.
Maybe something along these lines:

Group 1, build settings: Scala 2 / 3 toggle, sbt / Scala CLI toggle (by default: 3 & sbt)
Group 2, project name & group ID. Maybe we should have defaults here as well (such as: com.softwaremill / my-tapir - or sth auto-generated like curious-badger ;) ) so that generating a project can be even quicker if somebody just wants to explore?
Group 3: effect & server, if we would go with the something-chosen-by-default route, IO + http4s is a good choice
Group 4: options, swagger / json / metrics

adamw:
cc @lubarskyy

tpacuszkapl:
@adamw @lubarskyy - I've divided it into fieldsets, and added info icon. Let me know if that's what you had in mind:
![image](https://user-images.githubusercontent.com/79414144/188889822-d9fbe2dd-36e6-43b2-b2ce-bae9b30a8fa2.png)

adamw:
Looks good :) 

I'd only rename `Builder` to `Build tool`

lubarskyy:
Nice, I like it :+1: One nit pick, I would probably try to match the spacing between section header and text fields as they are not the same as section header to radio groups. Also I wonder how the section headers would look like if we apply the secondary color to them :thinking:

tpacuszkapl:
@adamw yes I have it, just I can't find time to polish it. But I think next week I will finish it.

adamw:
Ah ok thx :)

geminicaprograms:
> Looks good :)
> 
> I'd only rename `Builder` to `Build tool`

That could/should be actually done prior the change in question ;) - here is the PR #182

PR #182, opened 2022-09-15
feat: rename 'Builder' section to 'Build tool'

geminicaprograms:
So that it better reflects the intention behind the section.

Created as from comment to #130


adamw:
Thx :)

PR #183, opened 2022-09-15
feat: remove akka-http from the server options

### LINKED ISSUE #180: Remove akka-http from server options

geminicaprograms:
The `akka-http` option was removed from both backend and UI.

After this change there is no single reference to `akka` in the `adopt-tapir` anymore
```
> $ git --no-pager grep -in "akka" | wc -l
       0
```

Closes #180


adamw:
Thanks!

PR #188, opened 2022-09-20
feat: add label to distinguish generate ZIP/preview operations in metric

geminicaprograms:
At present preview is not reported as a metric. Add _operation_ label to the existing _adopt_tapir_starter_generated_total_ metric. It can contain the following values `preview` and `generate`. As a result both `preview` and `generate` operations can be independently tracked.

Part of #175

### DIFF
--- a/backend/src/main/scala/com/softwaremill/adopttapir/metrics/Metrics.scala
+++ b/backend/src/main/scala/com/softwaremill/adopttapir/metrics/Metrics.scala
@@ -5,7 +5,29 @@ import com.softwaremill.adopttapir.starter._
 import io.prometheus.client.{Counter, hotspot}
 
 object Metrics {
-  lazy val generatedStarterCounter: Counter =
+  def init(): Unit = hotspot.DefaultExports.initialize()
+
+  def increaseZipGenerationMetricCounter(details: StarterDetails): Unit = {
+    increaseMetricCounter(details, "generate")
+  }
+
+  def increasePreviewOperationMetricCounter(details: StarterDetails): Unit = {
+    increaseMetricCounter(details, "preview")
+  }
+
+  private def increaseMetricCounter(details: StarterDetails, operation: String): Unit = {
+    val labelValues = details.productElementNames
+      .zip(details.productIterator.toList)
+      .filterNot { case (name, _) => excludedStarterDetailsFields.contains(name) }
+      .map(_._2.toString)
+      .toList :+ operation
+
+    generatedStarterCounter
+      .labels(labelValues: _*)
+      .inc()
+  }
+
+  private lazy val generatedStarterCounter: Counter =
     Counter
       .build()
       .name(s"adopt_tapir_starter_generated_total")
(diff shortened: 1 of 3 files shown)


backend/src/main/scala/com/softwaremill/adopttapir/metrics/Metrics.scala
```diff
@@ -33,9 +53,9 @@ object Metrics {
         Builder.Sbt
       )
 
-    val names = fakeInstance.productElementNames.toArray.filterNot(excludedStarterDetailsFields.contains)
+    val names = fakeInstance.productElementNames.toArray.filterNot(excludedStarterDetailsFields.contains) :+ "operation"
     require(
-      names.length == fakeInstance.productElementNames.length - excludedStarterDetailsFields.size,
+      names.length == fakeInstance.productElementNames.length - excludedStarterDetailsFields.size + 1,
```
rafalambrozewicz, line 58: `1` => `OperationNameOffset`? Perhaps  `fieldNames` or`lables` would be more descriptive, than `names`? If so `OperationFieldNameOffset` or similar should be used.

backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterService.scala
```diff
@@ -26,23 +25,9 @@ class StarterService(projectGenerator: ProjectGenerator, filesManager: FilesMana
                 generatedFiles <- formattedGeneratedFiles
                 _ <- filesManager.createFiles(tempDir, generatedFiles)
                 zippedFile <- filesManager.zipDirectory(tempDir)
-                _ <- increaseMetricCounter(starterDetails)
+                _ <- IO(Metrics.increaseGenerationMetricCounter(starterDetails))
```
rafalambrozewicz, line 28: `increaseGenerationMetricCounter` => `increaseZipGenerationMetricCounter`? We could generate preview as well, so adding "zip" make this more explicit.

rafalambrozewicz (approved):
Looks good to me :-)

adamw (approved):
except the PR title ;)

backend/src/main/scala/com/softwaremill/adopttapir/metrics/Metrics.scala
geminicaprograms: I have introduced `additionalLabels` array that contains `operation` label (so far and it can be extended if needed later) and it can be used for both lines #56 and #58

backend/src/main/scala/com/softwaremill/adopttapir/starter/StarterService.scala
geminicaprograms: done, I have also renamed `increasePreviewMetricCounter` to `increasePreviewOperationMetricCounter` so that it is more explicit

geminicaprograms:
> except the PR title ;)

right it better reflects the intent of the change now

PR #191, opened 2022-09-23
divided fields into fieldsets; added initial data; added tooltip about server input dependencies

### LINKED ISSUE #130: Tidy up the main UI form

tpacuszkapl:
also increased the spacing between section header and text fields, and changed color of fieldset headers to secondary as suggested in #130.

closes #130


adamw:
thanks - some frontend tests seem to be failing though

tpacuszkapl:
Ah, right. I will fix it today.

tpacuszkapl:
I've fixed the tests. Also, the screen, because I forgot to add it before:
current

adamw:
Awesome, thanks! :)

PR #203, opened 2022-10-03
feat: enable 'preview' button in the embedded form

### LINKED ISSUE #169: Embedded adopt-tapir: add button to preview
The preview button should open the preview in a new tab, on the adopt-tapir page - so that we have enough space

geminicaprograms:
Note that clicking the `preview` button in the embedded form mode results in:
* generation of URL that contains the current configuration as query params with an extra `preview=true` parameter added
* opening of the URL in question in the new window
* configuration application followed by the switch to `preview` mode

Closes #169

### DIFF
--- a/ui/src/api/starter.ts
+++ b/ui/src/api/starter.ts
@@ -42,10 +42,9 @@ export type StarterRequest = {
   builder: Builder;
 };
 
+export const serverAddress = process.env.REACT_APP_SERVER_ADDRESS ?? 'https://adopt-tapir.softwaremill.com';
+
 export async function doRequestStarter(formData: StarterRequest) {
-  const serverAddress = !process.env.REACT_APP_SERVER_ADDRESS
-    ? 'https://adopt-tapir.softwaremill.com'
-    : process.env.REACT_APP_SERVER_ADDRESS;
   const response = await fetch(`${serverAddress}/api/v1/starter.zip`, {
     method: 'POST',
     headers: {
--- a/ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
+++ b/ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
@@ -1,4 +1,4 @@
-import { SyntheticEvent, useContext, useEffect, useState } from 'react';
+import { SyntheticEvent, useCallback, useContext, useEffect, useState } from 'react';
 import { Box, Button, IconButton, Stack, Tooltip, Typography } from '@mui/material';
 import ShareTwoToneIcon from '@mui/icons-material/ShareTwoTone';
 import { FormProvider, useForm } from 'react-hook-form';
@@ -46,7 +46,7 @@ export const ConfigurationForm: React.FC<ConfigurationFormProps> = ({ isEmbedded
   const initialData = useInitialData();
   const [initialized, setInitialized] = useState(false);
 
-  const [sharedRequest, snackbarConfig, ready] = useSharedConfig();
+  const [sharedRequest, snackbarConfig, ready, preview] = useSharedConfig();
 
   const form = useForm<StarterRequest>({
     mode: 'onBlur',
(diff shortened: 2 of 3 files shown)


ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
```diff
@@ -62,6 +69,23 @@ export const ConfigurationForm: React.FC<ConfigurationFormProps> = ({ isEmbedded
...
 
+  const handleShowPreview = useCallback(() => {
+    // 'trigger()' triggers the validation.
+    form.trigger().then(isValid => {
+      if (isValid) {
+        const casted = form.getValues() as StarterRequest;
+        // Conversion of bools is done by hand, because casting writes booleans as strings.
+        const formData: StarterRequest = {
+          ...casted,
+          addDocumentation: 'true' === casted.addDocumentation.toString(),
```
kkolasinski, line 80: tip: you can do `JSON.parse(casted.addDocumentation.toString())` to get the boolean `true` from a string if you're sure that it will always be either `'true'` or `'false'`

not sure about the readability of this approach tho :)

ui/src/api/starter.ts
```diff
@@ -42,10 +42,11 @@ export type StarterRequest = {
   builder: Builder;
 };
 
+export const serverAddress = !process.env.REACT_APP_SERVER_ADDRESS
+  ? 'https://adopt-tapir.softwaremill.com'
+  : process.env.REACT_APP_SERVER_ADDRESS;
```
kkolasinski, line 47: You can use the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) here, assuming you're using Node 14.0.0 or higher
```
export const serverAddress = process.env.REACT_APP_SERVER_ADDRESS ?? 'https://adopt-tapir.softwaremill.com'
```

kkolasinski (approved):

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
```diff
@@ -62,6 +69,23 @@ export const ConfigurationForm: React.FC<ConfigurationFormProps> = ({ isEmbedded
     },
   });
 
+  const handleShowPreview = useCallback(() => {
+    // 'trigger()' triggers the validation.
+    form.trigger().then(isValid => {
```
rafalambrozewicz, line 74: Is it possible to rename `trigger()` to `triggerTheValidation()` and then, thanks to that, remove the comment above?

ui/src/hooks/useSharedConfig.ts
```diff
@@ -5,28 +5,33 @@ import { StarterRequest } from '../api/starter';
...
-          setSnackbar({
-            open: true,
-            severity: 'info',
-            message: 'Linked configuration was applied.',
-          });
+          if (!shouldPreview) {
+            setSnackbar({
+              open: true,
+              severity: 'info',
+              message: 'Linked configuration was applied.',
```
rafalambrozewicz, line 32: AFAIK it is a good practice to separate the message's content from the code. Perhaps there is some component in the project acting like messages repository and this message should be placed there? If not, please ignore this suggestion.

rafalambrozewicz (commented):
Added some suggestions, looks ok to me, but my option is not particularly decisive in this case.

ui/src/api/starter.ts
geminicaprograms: done

ui/src/components/ConfigurationForm/ConfigurationForm.component.tsx
geminicaprograms: Not really as it is the part of the `react-hook-api`

ui/src/hooks/useSharedConfig.ts
geminicaprograms: I would say that we should discuss it as a separate issue for `adopt-tapir` app - if there are any plans to i18n it.

geminicaprograms: No plans ATM

Issue #194, opened 2022-09-27
Few tweaks to the structure generated for Scala CLI


romanowski:
I need to say that the project is amazing, and the generated structure for Scala CLI is good. Since we haven't created a recommended structure for such projects yet, adopt-tapir is the perfect opportunity to create it ;)

To kick off discussion, I created [a repo](https://github.com/romanowski/adopt-tapir-scala-cli-example/tree/master) where I track all the changes that I think should be made in the structure (if people agree I think I should be able to open PR to the tool changing the structure. Summary of change:
 - .gitignore mainly  to exclude `.scala-build` and `.bsp` directories
 - Use comment based using directives - other forms are not recommended for now and confuse Intellij / Metals. Metals nicely colors the comment based directives
 -  Use `*.scala` files in project definition not to confuse metals / IntelliJ
 - Add a bit to Readme about opening the project in IDE

geminicaprograms:
> 

Thanks for the feedback I really appreciate it 😃 
I generally agree and can perform the necessary changes in the code but I would leave the `.gitignore` out of the discussion for the time being as it is not generated for `sbt` either. IOW it is a feature itself and IMHO it should be better to introduce it for both `build tools`.

geminicaprograms:
Having said that I believe that we could join the efforts and I could prepare the PR (as I know the code better) and you could test it, @romanowski WDYT?

PR #210, opened 2022-10-04
feat: tune the generated Scala CLI project according to the guidelines

geminicaprograms:
The following suggestions were implemented:
* entry about importing Scala CLI project to popular IDEs (Metals / IntelliJ) was added to the generated `README.md`
* project files (`*.sc`) were renamed to `*.scala` so that IDEs are not confused, what is more `test.sc` was renamed to `build.test.scala` to better reflect its intent
* comment-based `using` notation is used in the generated project files as it is the recommended way of doing it

Note that `.gitignore` was not introduced as it is a good candidate for both `Build tools` and as such will be added as a part of separate PR.

Part of #194


adamw:
Thx :)

PR #214, opened 2022-10-05
fix: reduce buttons padding to 14px

geminicaprograms:
Issue: when embedded form is viewed under Safari it divides button into 2 lines.
image

Solution: reduce the default padding from `16px` to `14px` so that they fit into one line.
image

Follow-up to #169


lubarskyy (approved):
:1st_place_medal:

rafalambrozewicz (approved):

adamw:
Works great :)

PR #215, opened 2022-10-06
fix: don't show double vertical scrollbars on sbtx file

### LINKED ISSUE #192: Preview always get vertical scroll
The main window always has a scroll - and it gets larger when switching to the longer files (e.g. `Endpoints.scala`)

Safari/macos

image

geminicaprograms:
According to [SO](https://stackoverflow.com/a/61844861) it was determined that the top container has to have
fixed height (could be `100vh` related but cannot be `100%` related) and
then inner container can have 100% height. Note that `minHeight` was set to `450px`.

Apart from the the following changes were applied
* all margins were reduced so that file/code areas use more screen
  estate
* selected file (in file view) is no longer bolded (but still
  highlighted so that there is no doubt that it is selected) so that it
  doesn't take more space
* proportion between file and content views was modified so that more
  space is available for file view (potentially reducing the need for
  horizontal scrollbar)

Here is the current look & feel for `1794px` width:
image

Closes #92

### DIFF
--- a/ui/src/components/FileContentView/FileContentView.component.tsx
+++ b/ui/src/components/FileContentView/FileContentView.component.tsx
@@ -74,7 +74,6 @@ export function FileContentView({ files, opened }: Props) {
           backgroundColor: 'rgb(255, 255, 255)',
           padding: 0,
           height: '100%',
-          overflowY: 'auto',
         }}
         lineNumberStyle={{ minWidth: '44px' }}
         language={language}
--- a/ui/src/components/FileContentView/FileContentView.styles.ts
+++ b/ui/src/components/FileContentView/FileContentView.styles.ts
@@ -5,6 +5,7 @@ export const useStyles = makeStyles()(theme => {
     wrapper: {
       height: '100%',
       marginTop: 0,
+      overflow: 'auto',
     },
   };
 });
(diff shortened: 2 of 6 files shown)


adamw:
Great :) though I guess the no-scrollbar thing depends on the screen width?

geminicaprograms:
> Great :) though I guess the no-scrollbar thing depends on the screen width?

yeap, it will show-up anyway if the screen is small enough and we can't do much about it :/ the good thing is that by _default_ (in your and mine screens ;)) it will not show 😸

adamw:
Turns out my screen is 1792 wide ;-) 

Anyway, reading through my original issue, the problem is with the *vertical* scroll. Try opening `sbtx`, for example, and you get two vertical scrolls: one for the code viewer, and one for the entire window - the code viewer gets stretched. The second one should be there.

As for the changes here: the code is unfortunately too small now, it looks weird. So I would keep the current size. Let's keep the smaller margins - no reason not to use as much screen as we can.

geminicaprograms:
> Turns out my screen is 1792 wide ;-)
> 
> Anyway, reading through my original issue, the problem is with the _vertical_ scroll. Try opening `sbtx`, for example, and you get two vertical scrolls: one for the code viewer, and one for the entire window - the code viewer gets stretched. The second one should be there.
> 
> As for the changes here: the code is unfortunately too small now, it looks weird. So I would keep the current size. Let's keep the smaller margins - no reason not to use as much screen as we can.

I have reworked it so that `sbtx` file (and others too) are displayed correct. As suggested:
* kept the bigger screen estate to be utilised
* removed content view font size reduction

adamw:
Works great now, thanks :)

PR #224, opened 2022-10-13
fix: iframe height

### LINKED ISSUE #217: fix: integration with https://tapir.softwaremill.com/en/latest/#adopt-a-tapir
At present it requires a fixed height to be provided in the `iframe` definition e.g. `height="590"`. The problem is that the `iframe` itself has constraints that are not easily foreseeable (for instance on width) and it is hard to predict what would be the necessary size prior publishing the changes to UI. As a results we end up often in the situation like that:
image

where the content is clipped and vertical scrollbar is needed/show to see (sometimes like above they are partially visible) buttons.

If possible, provide the solution that doesn't require changes in the commit [#e868ea6f](https://github.com/softwaremill/tapir/commit/e868ea6fe9b869a276e3756e956135a1447fd442).

Note that iframe example would have to be modified in [README.md](/softwaremill/adopt-tapir/tree/main/ui#routing--embeddable-version)

marekhuckmann:
fixes #217

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -56,6 +56,6 @@ Example:
   sandbox="allow-scripts allow-same-origin allow-forms allow-downloads allow-popups"
   src={'https://adopt-tapir.softwaremill.com/embedded-form'}
   width="895"
-  height="590"
+  height="603"
 />
 ```


adamw:
Also needs to be fixed in `doc/index.md`

adamw:
Thx :)

PR #245, opened 2022-11-08
Fixing no spaces between sentences in generated code

rafalambrozewicz:
Added spaces in the `@if/ else` section of code templates, thus fixing the issue.


rafalambrozewicz:
At the first glance, it looks like a problem within templating library. It is fairly minor, however, perhaps it is worth investigating it a bit more and potentially raising an issue for twirl?

PR #277, opened 2023-01-01
Refactor templates for cats and zio

katlasik:
Refactor templates:
-  Use built-in methods for reading from and printing to the console in ZIO and cats. 
- Change the name of the env variable for setting the port number to HTTP_PORT.


adamw:
Thanks!

PR #279, opened 2023-01-01
Refactor CorrelationId

katlasik:
`CorrelationId` no longer is created using `unsafeRunSync`. It is now initialized in `Main` and passed down as an implicit parameter.


adamw:
As I understand the change is that we now pass the way to obtain a correlation id as an explicit dependency. Which is nice as it's not global, and the compiler gives an error when you don't have access to the implicit. On the other hand, it's a bit harder to use ... I wonder if this approach would carry over to bootzooka (on which adopt-tapir is based), where additionally we have a hack to use correlation ids from within `ConnectionIO`s.

adamw:
@katlasik could you create an issue in bootzooka to port or at least attempt to port this over, and/or a  PR?

katlasik:
Sure, I wil create an issue and then try to implement it.

PR #280, opened 2023-01-02
Refactor to use IO app

katlasik:
Refactor application to use IOApp. Replace explicit throws with Either.

### DIFF
--- a/backend/src/main/scala/com/softwaremill/adopttapir/Main.scala
+++ b/backend/src/main/scala/com/softwaremill/adopttapir/Main.scala
@@ -1,27 +1,15 @@
 package com.softwaremill.adopttapir
 
-import cats.effect.IO
-import cats.effect.unsafe.implicits.global
+import cats.effect.{ExitCode, IO, IOApp, Resource}
 import com.softwaremill.adopttapir.config.Config
-import com.softwaremill.adopttapir.metrics.Metrics
-import com.typesafe.scalalogging.StrictLogging
 
-object Main extends StrictLogging:
-  def main(args: Array[String]): Unit = {
-    Metrics.init()
-    Thread.setDefaultUncaughtExceptionHandler((t, e) => logger.error("Uncaught exception in thread: " + t, e))
+object Main extends IOApp:
 
-    val config = Config.read
-    Config.log(config)
+  override def run(args: List[String]): IO[ExitCode] =
+    val server = for
+      config <- Resource.eval(Config.read)
+      dependencies <- Dependencies.wire(config)
+      _ <- dependencies.api.resource
+    yield ExitCode.Success
 
-    Dependencies
-      .wire(config)
-      .use { case Dependencies(httpApi) =>
-        /*
-        - allocates the http api resource, and never releases it (so that the http server is available
-          as long as our application runs)
-         */
-        httpApi.resource.use(_ => IO.never)
-      }
-      .unsafeRunSync()
-  }
+    server.useForever
--- a/backend/src/main/scala/com/softwaremill/adopttapir/Main.scala
+++ b/backend/src/main/scala/com/softwaremill/adopttapir/Main.scala
@@ -2,11 +2,13 @@ package com.softwaremill.adopttapir
 
 import cats.effect.{ExitCode, IO, IOApp, Resource}
 import com.softwaremill.adopttapir.config.Config
+import com.softwaremill.adopttapir.infrastructure.CorrelationId
 
 object Main extends IOApp:
 
   override def run(args: List[String]): IO[ExitCode] =
     val server = for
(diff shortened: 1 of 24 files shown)


geminicaprograms:
👍 yet I would let the others to finally approve it 😄

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/DirectoryMerger.scala
```diff
@@ -1,44 +1,44 @@
...
-  def apply(topDirName: String, pathNames: List[String], content: String): Directory =
-    assert(pathNames.nonEmpty, "`pathNames` list cannot be empty!")
+object DirectoryMerger:
 
-    val pathsWithRootDir = List(topDirName) ++ pathNames
-    createTreeRec(pathsWithRootDir, content) match {
-      case File(_, _)          => throw new IllegalStateException("Top element of created tree cannot be a `File`!")
-      case d @ Directory(_, _) => d
+  def apply(topDirName: String, pathNames: List[String], content: String): Either[IllegalStateException, Directory] = for
+    _ <- Either
```
adamw, line 8: Hm I'm not sure about this syntax. Isn't a simple `if` + `Left`/`Right` simpler to read here? We wouldn't have to lift everything to a for-comprehension

backend/src/main/scala/com/softwaremill/adopttapir/Main.scala
```diff
@@ -1,27 +1,17 @@
...
 
-object Main extends StrictLogging:
-  def main(args: Array[String]): Unit = {
-    Metrics.init()
-    Thread.setDefaultUncaughtExceptionHandler((t, e) => logger.error("Uncaught exception in thread: " + t, e))
+object Main extends IOApp:
 
-    val config = Config.read
-    Config.log(config)
+  override def run(args: List[String]): IO[ExitCode] =
```
adamw, line 9: much nicer :)

backend/src/main/scala/com/softwaremill/adopttapir/template/BuildView.scala
```diff
@@ -91,18 +93,18 @@ abstract class BuildView:
...
+      case ServerEffectAndImplementation(FutureEffect, Netty) => HttpDependencies.netty().asRight
+      case ServerEffectAndImplementation(FutureEffect, VertX) => HttpDependencies.vertX().asRight
+      case ServerEffectAndImplementation(IOEffect, Http4s)    => HttpDependencies.http4s().asRight
+      case ServerEffectAndImplementation(IOEffect, Netty)     => HttpDependencies.ioNetty().asRight
+      case ServerEffectAndImplementation(IOEffect, VertX)     => HttpDependencies.ioVerteX().asRight
+      case ServerEffectAndImplementation(ZIOEffect, Http4s)   => HttpDependencies.http4sZIO().asRight
+      case ServerEffectAndImplementation(ZIOEffect, ZIOHttp)  => HttpDependencies.ZIOHttp().asRight
+      case ServerEffectAndImplementation(ZIOEffect, VertX)    => HttpDependencies.ZIOVerteX().asRight
+      case ServerEffectAndImplementation(ZIOEffect, Netty)    => HttpDependencies.ZIONetty().asRight
+      case other: StarterDetails => UnsupportedOperationException(s"Cannot pick dependencies for $other").asLeft
```
adamw, line 107: maybe we could get rid of the `Either` (which propagates quite deeply), by making sure that we can't create incorrect `StarterDetails` in the first place?

adamw: Then in the `other` branch we could safely throw an exception as this would be a programming bug, not  a user error

adamw:
Similarly as before, it would be great to propagate some of these changes to bootzooka

backend/src/main/scala/com/softwaremill/adopttapir/template/BuildView.scala
katlasik: Wouldn't it be unsafe to throw an error outside the effect system? This way (with Either converted to IO), it's passed to IO and only kills a fiber. So maybe I could just return `IO` from here.  My idea was to return `Either` to point out this is pure, not side-effecting code, but it can fail. 

So I could move this part of the validation to FormValidator, but then I think it would still be unsafe to throw an exception here (unless I wrap the whole function in `IO.apply` as it was done before).

I was also thinking to change this part to use `PartialFunction`, but I don't think it will be less bloated than Either.

rucek: How about representing every valid combination of a server and an effect as a value in a sealed ADT, so that illegal combinations are simply not representable? Then the `case other` won't be necessary at all.

PR #305, opened 2023-01-26
Change publish local to publish in build

PanHNE:
In build.sbt I changed 
* Docker / publishLocal := (Docker / publishLocal).dependsOn(copyWebapp).value
to 
* Docker / publish := (Docker / publish).dependsOn(copyWebapp).value

Also I added "-Dsbt.task.timings=true" to baskend-start.sh to get time report

### DIFF
--- a/backend-start.sh
+++ b/backend-start.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-sbt "~backend/reStart"
+sbt -Dsbt.task.timings=true "~backend/reStart"
--- a/backend-start.sh
+++ b/backend-start.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-sbt -Dsbt.task.timings=true "~backend/reStart"
+sbt "~backend/reStart"
--- a/build.sbt
+++ b/build.sbt
@@ -159,7 +159,7 @@ lazy val dockerSettings = Seq(
   Docker / packageName := "adopttapir",
   dockerUsername := Some("softwaremill"),
   dockerUpdateLatest := true,
-  Docker / publish := (Docker / publish).dependsOn(copyWebapp).value,
+  Docker / publishLocal := (Docker / publishLocal).dependsOn(copyWebapp).value,
   Docker / version := git.gitDescribedVersion.value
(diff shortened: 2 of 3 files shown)


backend-start.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-sbt "~backend/reStart"
+sbt -Dsbt.task.timings=true "~backend/reStart"
```
tpacuszkapl, line 3: I think it should be added to `Publish docker image` step in `/github/workflows/adopt-tapir-ci.yml` as we want to time the tasks during the deployment process.

build.sbt
```diff
@@ -159,7 +159,7 @@ lazy val dockerSettings = Seq(
   Docker / packageName := "adopttapir",
   dockerUsername := Some("softwaremill"),
   dockerUpdateLatest := true,
-  Docker / publishLocal := (Docker / publishLocal).dependsOn(copyWebapp).value,
+  Docker / publish := (Docker / publish).dependsOn(copyWebapp).value,
```
tpacuszkapl, line 162: We probably still want for `publishLocal` to depend on `copyWebapp`. Otherwise we won't get webapp in images created but not published to remote.

backend-start.sh
PanHNE: Ok, I will move -Dsbt.task.timings=true from backend-start.sh to adopt-tapir-ci.yml

build.sbt
tpacuszkapl, line 162: I've taken a look at the `Docker / publish` dependencies, and it seems it still depends on `publishLocal`:
`[info] Dependencies:
[info]  backend / Docker / dockerExecCommand
[info]  backend / Docker / dockerAliases
[info]  backend / Docker / dockerBuildxPlatforms
[info]  backend / Docker / publishLocal
[info]  backend / Docker / dockerBuildOptions
[info]  backend / Docker / publish / streams
[info]  backend / Docker / stage
`
So additional `dependOn` probably won't work. I would move it a little higher, so f.e. compile java or build jar tasks should depend on the `copyWebapp`.

PR #360, opened 2023-04-28
Update tapir

PanHNE:


kciesielski (approved):

backend/src/main/scala/com/softwaremill/adopttapir/template/BuildView.scala
```diff
@@ -134,7 +135,8 @@ abstract class BuildView:
 
     def ioVerteX(): List[ScalaDependency] = List(
       ScalaDependency("com.softwaremill.sttp.tapir", "tapir-vertx-server-cats", getTapirVersion()),
-      ScalaDependency("com.softwaremill.sttp.tapir", "tapir-cats", getTapirVersion())
+      ScalaDependency("com.softwaremill.sttp.tapir", "tapir-cats", getTapirVersion()),
```
adamw, line 138: is the -cats dependency needed?

PanHNE: I checked and we don't needed

kciesielski: I suggested to keep it, because I thought it was pulled by `tapir-cats-effect` transitively and I thought this way we leave the template project with a better visibility of dependencies. However, I just checked and they are indeed independent, and both use `cats-core`, so we can indeed remove `tapir-cats`.

PR #401, opened 2023-06-21
Set cpuStarvationCheckInterval for 5 seconds

PanHNE:


backend/src/main/scala/com/softwaremill/adopttapir/Main.scala
```diff
@@ -3,9 +3,14 @@ package com.softwaremill.adopttapir
 import cats.effect.{ExitCode, IO, IOApp, Resource}
 import com.softwaremill.adopttapir.config.Config
 import com.softwaremill.adopttapir.infrastructure.CorrelationId
+import cats.effect.unsafe.IORuntimeConfig
+import scala.concurrent.duration._
 
 object Main extends IOApp:
 
+  override def runtimeConfig: IORuntimeConfig =
+    super.runtimeConfig.copy(cpuStarvationCheckInterval = 5.seconds)
```
adamw, line 12: maybe let's do something less radical, the default is 100ms? so maybe let's try 500ms, what do you think? 5 seconds seems like entirely disabling the detector :)

kciesielski (approved):

PR #402, opened 2023-06-22
Set cpuStarvationCheckInterval for 5 seconds

PanHNE:


adamw:
but clearly the 500ms didn't take effect, what would it help to increase this to 5s?

PR #475, opened 2023-10-05
fix: `preview` metrics are not emitted

### LINKED ISSUE #474: [backend]: ContentServiceTest doesn't test anything
**Problem**
The [ContentServiceTest](https://github.com/softwaremill/adopt-tapir/blob/main/backend/src/test/scala/com/softwaremill/adopttapir/starter/content/ContentServiceTest.scala#L27) walks over all valid configurations, allocates a resource but doesn't eventually calls it (e.g. with `unsafeRunSync` or similar).

**Solution**
Add call to `unsafeRunSync` to actually call the content generation. However all configurations are tested in [StarterSeviceITTest](https://github.com/softwaremill/adopt-tapir/blob/main/backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITTest.scala) hence there is no need to repeat it again. Therefore it should be reduced to run a single configuration.

geminicaprograms:
The following PR contains two fixes:
1. it fixes the `ContentServiceTest` as it was not calling the content service but was just creating service resource - note that the test gets reduced to a single scenario as so that verification is not doubled
2. it fixes the missing `preview` metrics issue - in addition it add the unit tests that verifies if the both content and generate services are emitting metrics

Closes #474

### DIFF
--- a/backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
+++ b/backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
@@ -21,5 +21,5 @@ final case class ContentService(generatedFilesFormatter: GeneratedFilesFormatter
         DirectoryMerger(projectName, paths, content)
       }))
       projectAsTree <- IO(projectAsDirTrees.reduce(DirectoryMerger.apply))
-      _ <- IO(Metrics.increasePreviewOperationMetricCounter(starterDetails))
+      _ <- Metrics.increasePreviewOperationMetricCounter(starterDetails)
     yield projectAsTree
--- a/backend/src/test/scala/com/softwaremill/adopttapir/starter/MetricsEmittedTest.scala
+++ b/backend/src/test/scala/com/softwaremill/adopttapir/starter/MetricsEmittedTest.scala
@@ -1,0 +1,54 @@
+package com.softwaremill.adopttapir.starter
+
+import cats.effect.{IO, Resource}
+import com.softwaremill.adopttapir.infrastructure.CorrelationId
+import com.softwaremill.adopttapir.metrics.Metrics
+import com.softwaremill.adopttapir.starter.content.ContentService
+import com.softwaremill.adopttapir.starter.files.{FilesManager, StorageConfig}
+import com.softwaremill.adopttapir.starter.formatting.GeneratedFilesFormatter
+import com.softwaremill.adopttapir.test.BaseTest
+import org.scalatest.Assertion
+
+class MetricsEmittedTest extends BaseTest:
+  import MetricsEmittedTest.*
+  import cats.effect.unsafe.implicits.global
+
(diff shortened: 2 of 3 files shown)


backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
```diff
@@ -21,5 +21,5 @@ final case class ContentService(generatedFilesFormatter: GeneratedFilesFormatter
         DirectoryMerger(projectName, paths, content)
       }))
       projectAsTree <- IO(projectAsDirTrees.reduce(DirectoryMerger.apply))
-      _ <- IO(Metrics.increasePreviewOperationMetricCounter(starterDetails))
+      _ <- Metrics.increasePreviewOperationMetricCounter(starterDetails)
```
adamw, line 24: ha that might be the reason why the number of 200 requests didn't match the metrics - we never counted previews ;)

adamw: good to have a test :)

kciesielski:
Nice catch :)

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
geminicaprograms: yeah I was wondering about the same :D

kciesielski: BTW shouldn't this emit a compiler warning?

kciesielski (approved):

backend/src/main/scala/com/softwaremill/adopttapir/starter/content/ContentService.scala
geminicaprograms: yeah - see the following
```
$ sbt clean compile | grep ContentService -c
0
```

I guess that in order to achieve it we would have to return sth from it and then ignore it at level uo...

geminicaprograms: On the bright side we have a test for it now 😄

adamw: `-Wvalue-discard` I guess should do the trick ... we don't have it by default? ;)

adamw: it's there (`show scalacOptions`). But that's a different case, we're not casting anything to unit, we simply created an `IO[IO[...]]` and explicitly discarded the result (`_ <- ...`). Probably uncatchable by the compiler

kciesielski: Yeah, I meant exactly this. At first - "we should be able to catch this with compiler flags!". But then - "we are doing `_ <- ...`, no hope there".

PR #479, opened 2023-10-06
Fix compilation warnings

### LINKED ISSUE #478: [backend] clean-up all compiler warnings

geminicaprograms:
The following warnings are fixed:
* about unused imports and parameters
* about discarded non-Unit value of type
* about `apirStaticContentEndpoints is deprecated`

_Note_
Warnings about `unused import` from Twirl generated sources e.g.
```
[warn] /Users/jcentkowski/workspace/sml/adopt-tapir/backend/src/main/twirl/Endpoints.scala.txt:1:1: unused import
[warn] @(
[warn] ^
```
were not fixed.
The problem is that regardless of what is written in [scala.lang.org](https://www.scala-lang.org/2021/01/12/configuring-and-suppressing-warnings.html) neither `src` nor `origin` nor `site` filters are available as of `scalac` version `3.3.1` - here is the output of `help` command
```
$ scalac -Wconf:help
Configure compiler warnings.
Syntax: -Wconf:<filters>:<action>,<filters>:<action>,...
multiple <filters> are combined with &, i.e., <filter>&...&<filter>

<filter>
  - Any message: any

  - Message categories: cat=deprecation, cat=feature, cat=unchecked

  - Message content: msg=regex
    The regex need only match some part of the message, not all of it.

  - Message id: id=E129
    The message id is printed with the warning.

  - Message name: name=PureExpressionInStatementPosition
    The message name is printed with the warning in verbose warning mode.

In verbose warning mode the compiler prints matching filters for warnings.
Verbose mode can be enabled globally using `-Wconf:any:verbose`, or locally
using the @nowarn annotation (example: `@nowarn("v") def test = try 1`).

<action>
  - error / e
  - warning / w
  - verbose / v (emit warning, show additional help for writing `-Wconf` filters)
  - info / i    (infos are not counted as warnings and not affected by `-Werror`)
  - silent / s

The default configuration is empty.

User-defined configurations are added to the left. The leftmost rule matching
a warning message defines the action.

Examples:
  - change every warning into an error: -Wconf:any:error
  - silence deprecations: -Wconf:cat=deprecation:s

Note: on the command-line you might need to quote configurations containing `*` or `&`
to prevent the shell from expanding patterns.
```

The only way to silence `unused import` warning was to add
```scala
scalacOptions += "-Wconf:msg=unused import:s",
```
to `build.sbt` but that would also silence warnings in every `*.scala` files hence it would be too much 😢 

Closes #478

### DIFF
--- a/backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITTest.scala
+++ b/backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITTest.scala
@@ -11,7 +11,6 @@ import com.softwaremill.adopttapir.metrics.Metrics
 import com.softwaremill.adopttapir.starter.api.*
 import com.softwaremill.adopttapir.starter.files.{FilesManager, StorageConfig}
 import com.softwaremill.adopttapir.starter.formatting.GeneratedFilesFormatter
-import com.softwaremill.adopttapir.template.ProjectGenerator
 import com.softwaremill.adopttapir.test.ServiceTimeouts.waitForPortTimeout
 import com.softwaremill.adopttapir.test.ShowHelpers.*
 import com.softwaremill.adopttapir.test.{BaseTest, GeneratedService, ServiceFactory}
@@ -21,7 +20,6 @@ import sttp.client3.{HttpURLConnectionBackend, Identity, SttpBackend, UriContext
 import scala.collection.mutable
 import scala.concurrent.duration.{DurationInt, FiniteDuration}
 import scala.util.Properties
-import scala.reflect.ClassTag
 
 object Setup:
   type TestFunction = Integer => IO[Unit]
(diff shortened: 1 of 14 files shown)


backend/src/test/scala/com/softwaremill/adopttapir/starter/StarterServiceITTest.scala
```diff
@@ -199,5 +197,5 @@ object ZipGenerator:
     yield service
 
 case class RunLogger(log: mutable.StringBuilder = new mutable.StringBuilder()):
-  def log(l: String): IO[Unit] = IO(log.append(System.lineSeparator()).append(l))
+  def log(l: String): IO[Unit] = IO(log.append(System.lineSeparator()).append(l).asInstanceOf[Unit])
```
kciesielski, line 200: **suggestion:** This look suspicious. What's the type of `.append(l)`? I think it would be a bit nicer to call `.void` on `IO`:
```scala
IO(log.append(System.lineSeparator()).append(l)).void
```

geminicaprograms: The result type of `append` is `StringBuilder` as it follows fluent API but I have counted on better suggestions 😸 🙇

adamw:
Thanks :)

Issue #471, opened 2023-10-02
Add Pickler as a JSON variant


kciesielski:
As now our [Pickler](https://tapir.softwaremill.com/en/latest/endpoint/pickler.html?highlight=pickler) feature is released, we'd like to showcase it in adopt-tapir.

PR #483, opened 2023-10-09
feat: add Pickler as json variant

DybekK:
Closes: https://github.com/softwaremill/adopt-tapir/issues/471 

This PR adds [Pickler](https://tapir.softwaremill.com/en/latest/endpoint/pickler.html) to json implementations only for `Scala 3`.  Changes include both unit and integration tests.

UI screenshot:

Screenshot 2023-10-09 at 20 43 21


kciesielski:
Nice, thanks!

PR #598, opened 2024-04-29
Support netty-sync

kciesielski:
This PR adds the `tapir-netty-sync` backend. It appears as a separate `Sync` effect type, available only for `Netty` and `Scala 3`. This backend has no "effect" per se, it's the direct-style monadless one, based on Netty and Loom. In order to test this properly, I added a new CI matrix axis for the effect. As a result, quite a lot of parallel builds are generated, but the overall build time doesn't increase (it's actually faster: ~13m instead of ~25m).
Additionally, I bumbed JDK to 21 in the workflow, leaving Java 17 for integration tests with effect != "Sync".

### DIFF
--- a/ui/src/components/ConfigurationForm/ConfigurationForm.consts.ts
+++ b/ui/src/components/ConfigurationForm/ConfigurationForm.consts.ts
@@ -37,6 +37,10 @@ export const EFFECT_TYPE_OPTIONS: FormSelectOption<EffectType>[] = [
     value: EffectType.IO,
   },
   {
+    label: 'Sync (direct-style)',
+    value: EffectType.Sync,
+  },
+  {
     label: 'ZIO',
     value: EffectType.ZIO,
   },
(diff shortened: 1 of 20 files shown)


ui/src/components/ConfigurationForm/ConfigurationForm.consts.ts
```diff
@@ -36,6 +36,10 @@ export const EFFECT_TYPE_OPTIONS: FormSelectOption<EffectType>[] = [
     label: 'IO (cats-effect)',
     value: EffectType.IO,
   },
+  {
+    label: 'Sync (direct-style)',
```
adamw, line 40: maybe let's label this as `No effect (direct style)`? As it's a dropdown value for "Effect type". 

Or maybe it would be best to rename the whole dropdown. Maybe "Stack"? With values: `Functional (IO, cats-effect)`, `Functional (ZIO)`, `Direct-style (Ox)`, `Future`? WDYT?

kciesielski: I like the Stack idea. This would mean quite a refactor to make the code reflect it accurately in all layers (StackType instead of EffectType, etc.), so maybe in the next PR?

adamw: Ah, I was just thinking of changing the labels, but renaming in code is probably the right thing to do ;)

adamw (approved):

PR #629, opened 2024-07-10
Update defaults to Ox+Netty+ScalaCLI

micossow:
Also auto-select implementation when stack is changed
Also updates scala 3 version to 3.4.2


adamw:
We should also probably remove the pin from https://github.com/softwaremill/adopt-tapir/blob/main/.scala-steward.conf

PR #635, opened 2024-07-12
Use OxApp for Netty sync server

kciesielski:


kciesielski:
@adamw Can I get a pre-review? It's still in draft mode because it has direct ox 0.3.1 dependency until Tapir catches up.

adamw:
But looks good :)

adamw:
Hm that's weird, steward even updates tapir deps in tapir itself (in the examples)

PR #940, opened 2026-04-22
Change metrics from prometheus to opentelemetry

kubinio123:
DONE:
* replace prometheus metrics and `/metrics` scrape endpoint with otel metrics and otlp push based exporter
* update integration tests to verify metrics export by spinning up stub collector server that awaits first metrics push

### DIFF
--- a/backend/src/main/twirl/MainFutureNetty.scala.txt
+++ b/backend/src/main/twirl/MainFutureNetty.scala.txt
@@ -12,7 +12,7 @@ object Main {
   def main(args: Array[String]): Unit = {
     @if(addMetrics) {
     val serverOptions = NettyFutureServerOptions.customiseInterceptors
-      .metricsInterceptor(Endpoints.prometheusMetrics.metricsInterceptor())
+      .metricsInterceptor(Endpoints.otelMetrics.metricsInterceptor())
       .options
     }
     val port = sys.env.get("HTTP_PORT").flatMap(_.toIntOption).getOrElse(8080)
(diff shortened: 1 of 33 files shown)


backend/src/main/twirl/MainFutureNetty.scala.txt
```diff
@@ -12,7 +12,7 @@ object Main {
   def main(args: Array[String]): Unit = {
     @if(addMetrics) {
     val serverOptions = NettyFutureServerOptions.customiseInterceptors
-      .metricsInterceptor(Endpoints.prometheusMetrics.metricsInterceptor())
+      .metricsInterceptor(Endpoints.otelMetrics.metricsInterceptor())
```
adamw: metrics is no longer an endpoint, right? so maybe this shouldn't be under `Endpoints`?

kubinio123: Right, but they are still only _for_ endpoints, via `tapir-opentelemetry-metrics` that comes with some basic ones. That's why I decided to keep it in the `Endpoints` file instead of a new file, keeps it simple.

adamw: hm yeah this might make sense ... but maybe just initialize in `Main` then? wouldn't that be simpler?

kubinio123: Like, just this line in `Main` 
```
val otel: OpenTelemetry = AutoConfiguredOpenTelemetrySdk.initialize().getOpenTelemetrySdk
``` 
then it wouldn't be simpler as that would need to passed to `Endpoints` to encapsulate all endpoints related stuff there. 

Btw. `OpenTelemetryMetrics` name is a little confusing when I think of it, it should be called `EndpointMetrics` maybe or `EndpointOpenTelemetryMetrics`. But that's not this repository :)

PR #945, opened 2026-04-24
Allow agents to use adopt tapir to generate project skeletons

### LINKED ISSUE #942: Allow agents to use adopt-tapir to generate project skeletons
What's the best approach here? Ideally I'd want to say sth like: "use adopt-tapir.softwaremill.com to generate a template scala project"

Do we need an agents.md / llms.txt to make this work? Expose an API endpoint? Or implement an MCP server? Though, adding an mcp server to your config just to generate a skeleton seems an overkill. In the end, the agent should just call the API and obtain a couple of files

kubinio123:
DONE:
* More textual OpenAPI description of `StarterRequest` properties with few common stack examples
* Add `llms.txt` file describing the app purpose, including links to Swagger UI and downloadable OpenAPI yaml, based on format https://llmstxt.org/#format

### DIFF
--- a/backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequest.scala
+++ b/backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequest.scala
@@ -5,16 +5,64 @@ import io.circe.generic.semiauto.*
 import io.circe.{Decoder, Encoder}
 import org.latestbit.circe.adt.codec.*
 import sttp.tapir.Schema
+import sttp.tapir.Schema.annotations.description
 
 case class StarterRequest(
+    @description("The name of the generated project")
     projectName: String,
+    @description("The group id for the generated project, e.g. com.softwaremill")
     groupId: String,
+    @description(
+      """The effect stack for the generated project. Each stack supports a specific set of server implementations:
+        |<ul>
+        |  <li><b>FutureStack</b>: Netty, VertX, Pekko</li>
+        |  <li><b>IOStack</b>: Netty, VertX, Http4s</li>
+        |  <li><b>ZIOStack</b>: Netty, VertX, Http4s, ZIOHttp</li>
+        |  <li><b>OxStack</b>: Netty</li>
+        |</ul>""".stripMargin
+    )
     stack: StackRequest,
+    @description(
+      """The HTTP server implementation to use. Must be compatible with the chosen <b>stack</b>:
+        |<ul>
+        |  <li><b>Netty</b></li>
+        |  <li><b>Http4s</b></li>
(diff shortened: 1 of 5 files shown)


backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequest.scala
```diff
@@ -5,16 +5,64 @@ import io.circe.generic.semiauto.*
...
 
 case class StarterRequest(
+    @description("The name of the generated project")
     projectName: String,
+    @description("The group id for the generated project, e.g. com.softwaremill")
     groupId: String,
+    @description(
+      """The effect stack for the generated project. Each stack supports a specific set of server implementations:
+        |<ul>
+        |  <li><b>FutureStack</b>: Netty, VertX, Pekko</li>
```
adamw: aren't we duplicating the enum values here? the enums are also part of the yaml after all

adamw:
Does it work? ;)

backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequest.scala
kubinio123: We are but I though a textual description could be a bonus for the llm too. Plus the enums doesn't reflect the validation of stack + server implementation.

kubinio123:
> Does it work? ;)

I don't know yet, merge to main will trigger the deploy and we'll see.

adamw:
If you start it locally and tell the AI to use adopt tapir @ localhost, wouldn't it be a good preliminary test?

kubinio123:
Ye, it would probably, let me try that.

backend/src/main/scala/com/softwaremill/adopttapir/starter/api/StarterRequest.scala
adamw: Hm well let's try without. Always less duplication :)

kubinio123:
Running locally I had to point claude to `llms.txt` it didn't check it upfront, then he did great from that point. Read the OpenAPI yaml from the url from llms.txt so the live app and prepared a questionare allowing me to choose between the options. Then fetched the zip. Additions to the endpoints docs are not needed I suppose, I will just keep the examples.

adamw:
Maybe we need some kind of link to llms.txt on the page? maybe in the header (as in `<link ...>`)? Is there some at least semi-standard way to do this?

kubinio123:
Hmm no standards yet from what I see, but putting those hints in some places does the trick; sometimes agent asks for `index.html` in the first place while sometimes just for server response headers. I think we can put the hint in both places to improve the visibility.

